PR #7, opened 2021-12-01
feature: add ability to export a json schema for opslevel.yaml

rocktavious:
Add the command `opslevel export schema` which emits a json schema of the structure which mirrors opslevel.yml

### DIFF
--- a/src/cmd/opslevel.go
+++ b/src/cmd/opslevel.go
@@ -1,0 +1,64 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/alecthomas/jsonschema"
+	"github.com/spf13/cobra"
+)
+
+type OpsLevelTag struct {
+	Key   string `json:"key"`
+	Value string `json:"value"`
+}
+
+type OpsLevelTool struct {
+	Name        string `json:"name"`
+	Category    string `json:"category"` // TODO: this should be an enum
+	Url         string `json:"url"`
+	Environment string `json:"environment,omitempty"`
+}
+
+type OpsLevelRepository struct {
+	Name     string `json:"name"`
+	Path     string `json:"path,omitempty"`
+	Provider string `json:"provider"` // TODO: this should be an enum
+}
+
+type OpsLevelService struct {
+	Name         string               `json:"name"`
+	Description  string               `json:"description"`
+	Owner        string               `json:"owner"`
+	Lifecycle    string               `json:"lifecycle"`
+	Tier         string               `json:"tier"`
+	Product      string               `json:"product"`
+	Language     string               `json:"language"`
+	Framework    string               `json:"framework"`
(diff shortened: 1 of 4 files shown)


src/cmd/opslevel.go
```diff
@@ -0,0 +1,69 @@
...
+	Path     string `json:"path,omitempty"`
+	Provider string `json:"provider"` // TODO: this should be an enum
+}
+
+type OpsLevelServiceDependancy struct {
+	Alias string `json:"alias"`
+}
+
+type OpsLevelService struct {
+	Name         string                      `json:"name"`
```
maxim-opslevel, line 35: why no `omitempty` on this field?

src/cmd/opslevel.go
```diff
@@ -0,0 +1,69 @@
...
+	Name        string                `json:"name"`
+	Category    opslevel.ToolCategory `json:"category"`
+	Url         string                `json:"url"`
+	Environment string                `json:"environment,omitempty"`
+}
+
+type OpsLevelRepository struct {
+	Name     string `json:"name"`
+	Path     string `json:"path,omitempty"`
+	Provider string `json:"provider"` // TODO: this should be an enum
```
maxim-opslevel, line 27: You're missing

DisplayName

maxim-opslevel:
This works for a `Service` opslevel.yml but not a `Repository` [opslevel.yml](https://www.opslevel.com/docs/api/opslevel-yml/#repository). I think we need to allow for the creation of a schema for this as well.

src/cmd/opslevel.go
rocktavious: Name is required

src/cmd/opslevel.go
```diff
@@ -0,0 +1,69 @@
...
+	Aliases      []string                    `json:"aliases,omitempty"`
+	Tags         []OpsLevelTag               `json:"tags,omitempty"`
+	Tools        []OpsLevelTool              `json:"tools,omitempty"`
+	Repositories []OpsLevelRepository        `json:"repositories,omitempty"`
+	Dependencies []OpsLevelServiceDependancy `json:"dependencies,omitempty"`
+}
+
+type OpsLevelConfig struct {
+	Version int             `json:"version"`
+	Service OpsLevelService `json:"service"`
```
rocktavious, line 52: @maxim-opslevel i think we just add this here and it builds 1 schema that works for both
```
Repository OpsLevelRepository `json:"repository"`
```

But i guess that question becomes - "can you define both a repository and a service inside 1 Git repo inside the same opslevel.yml?"

src/cmd/opslevel.go
rocktavious: Added

src/cmd/opslevel.go
rocktavious: Added "one_of" requirement

rocktavious:
json schema is now available via https://app.opslevel.com/public/schemas/opslevel.schema.yml

PR #11, opened 2022-02-10
Add interactive functionality for importing checks

eochoa260:


src/cmd/check.go
```diff
@@ -92,31 +92,72 @@ type CheckCreateType struct {
 	Spec map[string]interface{}
 }
 
-func (self *CheckCreateType) resolveAliases() {
+func (self *CheckCreateType) resolveCategoryAliases(client *opslevel.Client) (error) {
```
eochoa260, line 95: Took a stab at refactoring these based on what I had to do for the integration one

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+		Label:    "{{ . }}?",
+		Active:   fmt.Sprintf("%s {{ .Name | cyan }}", promptui.IconSelect),
+		Inactive: "    {{ .Name | cyan }}",
+		Selected: fmt.Sprintf("%s {{ .Name | faint }}", promptui.IconGood),
+		Details: `
+		{{ "Aliases:" | faint }}	{{ .Aliases }}
+		{{ "Manager:" | faint }}	{{ .Manager.Name }}`,
+	}
+
+	noneValue := opslevel.Team{
```
eochoa260, line 116: Pulled this from a "how to prepend to list" that I found but open to feedback if there's a better way to add an object to the front of a list.

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+	templates := &promptui.SelectTemplates{
+		Label:    "{{ . }}?",
+		Active:   fmt.Sprintf("%s {{ .Name | cyan }}", promptui.IconSelect),
+		Inactive: "    {{ .Name | cyan }}",
+		Selected: fmt.Sprintf("%s {{ .Name | faint }}", promptui.IconGood),
+		Details: `
+		{{ "Type:" | faint }}	{{ .Type }}`,
+	}
+
+	n := 0
```
eochoa260, line 150: Same as above, open to feedback but based this on an example for updating a list in place so did the same here to filter the list

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	templates := &promptui.SelectTemplates{
+		Label:    "{{ . }}?",
+		Active:   fmt.Sprintf("%s {{ .Name | cyan }}", promptui.IconSelect),
+		Inactive: "    {{ .Name | cyan }}",
+		Selected: fmt.Sprintf("%s {{ .Name | faint }}", promptui.IconGood),
+	}
+
+	noneValue := opslevel.Filter{
```
eochoa260, line 80: Same as below

src/cmd/check.go
```diff
@@ -92,31 +92,72 @@ type CheckCreateType struct {
...
 	if item, ok := self.Spec["category"]; ok {
 		if value, ok := opslevel.Cache.TryGetCategory(item.(string)); ok {
 			delete(self.Spec, "category")
 			self.Spec["categoryId"] = value.Id.(interface{})
 		}
```
rocktavious, line 100: We should add error handling here - if the user specified a category in the yaml but its not a valid value we should raise an error

src/common/prompts.go
rocktavious: Its bad practice to mutate a list that you are iterating over.  I would instantiate a new list and then append items to it as you loop through the original list.  This remove the need for the var `n` and the increment in the loop as well as the slice assign on 157.

Also you can combind loops and a single if statement like this

```
filteredList := []opslevel.Integration{}
for _, value := range list; val.Type == "generic" {
   filteredList = append(newList, val)
}
```

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	templates := &promptui.SelectTemplates{
+		Label:    "{{ . }}?",
+		Active:   fmt.Sprintf("%s {{ .Name | cyan }}", promptui.IconSelect),
+		Inactive: "    {{ .Name | cyan }}",
+		Selected: fmt.Sprintf("%s {{ .Name | faint }}", promptui.IconGood),
+		Details: `
+		{{ "Aliases:" | faint }}	{{ .Aliases }}
+		{{ "Manager:" | faint }}	{{ .Manager.Name }}`,
```
rocktavious, line 113: maybe use `.Manager.email` - we intend to have e-mail be the "alias" of `User` resources since it has to be unique.

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+			n++
+		}
+	}
+	list = list[:n]
+
+	prompt := promptui.Select{
+		Label:     "Select Integration",
+		Items:     list,
+		Templates: templates,
+		Size:      len(list),
```
rocktavious, line 163: after testing this out on circular the list becomes huge and if your terminal is small it gets weird.  We should probably wrap all of these in a [min function](https://pkg.go.dev/math#Min) so if the list is 6 items the value is 6 but if the list is 20 items the value is capped at 10.

rocktavious (changes requested):
Great first pass just a couple of things to fix - let me know if you need any help!

src/cmd/check.go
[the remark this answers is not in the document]
eochoa260: Sounds good, I'll start trying to rework the logic and let you know if I have any questions

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+	list, err := client.ListFilters()
+	if err != nil {
+		return nil, err
+	}
+
+	templates := &promptui.SelectTemplates{
+		Label:    "{{ . }}?",
+		Active:   fmt.Sprintf("%s {{ .Name | cyan }}", promptui.IconSelect),
+		Inactive: "    {{ .Name | cyan }}",
+		Selected: fmt.Sprintf("%s {{ .Name | faint }}", promptui.IconGood),
```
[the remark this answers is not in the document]
eochoa260: That would be cool, I'll try to take a look at this after making the other changes.

src/common/prompts.go
```diff
@@ -0,0 +1,171 @@
...
+		Details: `
+		{{ "Aliases:" | faint }}	{{ .Aliases }}
```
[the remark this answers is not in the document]
eochoa260: Hmm, I'll have to test that out, it only had one alias on the one I was testing with but I'll add a couple more and see how it renders.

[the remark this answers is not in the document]
eochoa260: Just tested this out and it shows a space separated list: 
```Aliases:        [platform test1 test2]```

File: src/cmd/check.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (eochoa260) [reply src/cmd/check.go:108]: Seems like I'm duplicating some code here but added the else block to prompt when the provided value isn't found

File: src/common/prompts.go [hunk 10]
```diff
@@ -0,0 +1,171 @@
...
+
+	index, _, err := prompt.Run()
+	if err != nil {
+		return nil, err
+	}
+	return &list[index], nil
+}
+
+func PromptForLevels(client *opslevel.Client) (*opslevel.Level, error) {
+	list, err := client.ListLevels()
```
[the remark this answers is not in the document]
eochoa260: Still working on this one... trying to find an easy way to strip an element from a slice but I'm finding most suggestions are to write a function to do it.

PR #15, opened 2022-02-22
Eo/fix predicate export

eochoa260:


src/cmd/terraform.go
```diff
@@ -403,6 +403,15 @@ func flattenPredicate(key string, value *opslevel.Predicate) string {
   }
 `
 	if value != nil {
+		if value.Type == "matches_regex" {
+			config = `
+  %s {
+    type = "%s"
+	  value = <<-EOT
```
eochoa260, line 410: Spacing was weird so I had to add spaces to get it to look correct on the export.  Not sure if you know what could be causing that.

rocktavious, line 410: Hmm - Whats the reason for adding an if statement here?  i guess is there a reason you didn't make the escaping change to line 402?  I would think it should be valid no matter the predicate type.

rocktavious (changes requested):

src/cmd/terraform.go
eochoa260: I was mainly going off the ticket, it says in there to do this for type == match_regex so I figured we didn't want to do it for the other types.  I agree that it would be cleaner if it also worked for other types so that we can just update the original config definition though.

rocktavious: Thats fair - but in that case to reduce the duplication of the template i'd do just extract the specific changes we need - IE:
```
	config := `
  %s {
    type = "%s"
    value = %s
  }
`
	if value != nil {
		predicateValue := ""
		if value.Type == "matches_regex" {
			predicateValue = fmt.Sprint(`<<-EOT
%s
EOT`, value.Value)
		} else {
			predicateValue = fmt.Sprint(`"%s"`, strings.ReplaceAll(value.Value, "\"", "\\\""))
		}
		return templateConfig(config, key, value.Type, predicateValue)
	}
	return ""
```

but i'm not sure we actually need to have the if statement - i leave it up to you - if we do go without the if statement and "value" is always escaped multiline string we should make sure we do extensive testing of importing and exporting and using terraform to manage the checks with multiple different kinds of predicate types so we can ensure we are not breaking anything.

eochoa260: So I was actually just testing this and it seems to work fine with other types on the export and terraform plan but I haven't applied the terraform to see what the end state looks like.  Code looks cleaner without the if statement though so let me apply and make sure that the resources get created as expected.

eochoa260: Tested applying and it seems to be working as expected for other types too

PR #16, opened 2022-03-04
Eo/add opslevel version

eochoa260:


src/cmd/version.go
```diff
@@ -1,24 +1,104 @@
...
+		Version:  runtime.Version(),
+		Compiler: runtime.Compiler,
+		OS:       runtime.GOOS,
+		Arch:     runtime.GOARCH,
+	}
+}
+
+func getOpslevelVersion() OpslevelVersion {
+	// Need to update all of this when we switch over to resty client
+	apiUrl := viper.GetString("base-url") + "/api/ping"
```
eochoa260, line 64: Tried to get path.Join to work here but was getting this:  ```Get "https:///app.opslevel-staging.com/api/ping": http: no Host in request URLexit status 1```
Looked it up but not sure why it was adding an extra slash after the protocol.

rocktavious: Ya we need to sanitize this because our end user will be setting `base-url` and they might add a / at the end

lets do this
```
url := url.Parse(viper.GetString("base-url"))
url.Path = "/api/ping"
response, err := http.Get(url.String())
````

File: src/cmd/version.go [hunk 2]
```diff
@@ -1,24 +1,104 @@
...
+
+	return opslevelVersion
+}
 
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Print version information",
 	Long:  `Print version information`,
-	Run:   runVersion,
+	RunE:  runVersion,
```
rocktavious, line 89: ouu fancy - does this just do cobra checkErr automatically?

src/cmd/root.go
```diff
@@ -22,21 +22,24 @@ var rootCmd = &cobra.Command{
...
 	version = v
+	commit = commit
 	cobra.CheckErr(rootCmd.Execute())
 }
 
 func init() {
 	rootCmd.PersistentFlags().String("log-format", "TEXT", "overrides environment variable 'OPSLEVEL_LOG_FORMAT' (options [\"JSON\", \"TEXT\"])")
 	rootCmd.PersistentFlags().String("log-level", "INFO", "overrides environment variable 'OPSLEVEL_LOG_LEVEL' (options [\"ERROR\", \"WARN\", \"INFO\", \"DEBUG\"])")
 	rootCmd.PersistentFlags().String("api-url", "https://api.opslevel.com/graphql", "The OpsLevel API Url. Overrides environment variable 'OPSLEVEL_API_URL'")
+	rootCmd.PersistentFlags().String("base-url", "https://app.opslevel.com", "The OpsLevel App Url. Overrides environment variable 'OPSLEVEL_APP_URL'")
```
rocktavious, line 35: lets use `app-url` since the ENV VAR is `OPSLEVEL_APP_URL`

src/cmd/version.go
eochoa260: So this is what replicated had in their command and I updated it to that because Run doesn't accept a function that returns an error.

src/cmd/version.go
eochoa260: Oh nice, yeah that's better.  Just updated it.

src/cmd/version.go
rocktavious: Nice - we could reduce alot of code if we refactor for this - i'll cut a backlog ticket

src/cmd/version.go
```diff
@@ -1,24 +1,110 @@
...
+		Compiler: runtime.Compiler,
+		OS:       runtime.GOOS,
+		Arch:     runtime.GOARCH,
+	}
+}
+
+func getOpslevelVersion() OpslevelVersion {
+	// Need to update all of this when we switch over to resty client
+	url, err := url.Parse(viper.GetString("app-url"))
+	if err != nil {
```
rocktavious, line 66: replace this with `cobra.CheckErr` thats essentially what this does

src/cmd/version.go
```diff
@@ -1,24 +1,110 @@
...
+		fmt.Print(err.Error())
+		os.Exit(1)
+	}
+	url.Path = "/api/ping"
+	response, err := http.Get(url.String())
+	if err != nil {
```
rocktavious, line 72: same

src/cmd/version.go
```diff
@@ -1,24 +1,110 @@
...
+		fmt.Print(err.Error())
+		os.Exit(1)
+	}
+
+	responseData, err := ioutil.ReadAll(response.Body)
+	if err != nil {
```
rocktavious, line 78: same

rocktavious (approved):

PR #20, opened 2022-04-11
Filter out index==0 for level options

eochoa260:


src/common/prompts.go
```diff
@@ -51,8 +51,8 @@ func PromptForLevels(client *opslevel.Client) (*opslevel.Level, error) {
 	}
 
 	filteredList := []opslevel.Level{}
-	for _, val := range list {
-		if val.Alias != "beginner" {
+	for i, val := range list {
+		if i != 0 {
```
rocktavious, line 55: `val` is a "Level" object which has its own notions of index so it would be better to use that incase the levels are returned in non index order
```
if val.Index == 0 {
  continue
}
filteredList = append(filteredList, val)
```

rocktavious (changes requested):

src/common/prompts.go
eochoa260: Should we also be ordering this list in case it's not?  I'm assuming the prompt options would be out of order if that were the case?

rocktavious: @eochoa260 - YES!  that a great callout. https://pkg.go.dev/sort#Slice

eochoa260: Ok cool! Just updated it.

rocktavious (approved):
LGTM

PR #23, opened 2022-04-20
Add get, list, and get subcommands

eochoa260:


src/cmd/group.go
```diff
@@ -38,9 +39,161 @@ EOF
...
+			group, err = getClientGQL().GetGroupWithAlias(key)
+		}
+		cobra.CheckErr(err)
+		common.PrettyPrint(group)
+		common.WasFound(group.Id == nil, key)
+	},
+}
 
-// TODO: List Groups
+var getMembersCommand = &cobra.Command{
```
rocktavious, line 63: lets call this `getGroupMembersCommand` because its nested within the group subcommand

src/cmd/group.go
```diff
@@ -38,9 +39,161 @@ EOF
...
+		}
+		cobra.CheckErr(err)
+		members, err := group.Members(getClientGQL())
+		cobra.CheckErr(err)
+		common.PrettyPrint(members)
+		common.WasFound(group.Id == nil, key)
+	},
+}
+
+var getDescendantRepositoriesCommand = &cobra.Command{
```
rocktavious, line 86: same prefix with `getGroup...` along with all the others that are nested

src/cmd/group.go
```diff
@@ -38,9 +39,161 @@ EOF
...
+	Use:        "repositories ID|ALIAS",
+	Short:      "Get descendant repositories for a group",
```
rocktavious, line 88: We should probably use the descriptions in the API spec

In this case specifically
```
All the repositories that fall under this group - ex. this group's child repositories, all the child repositories of this group's descendants, etc.
```
Maybe just for the Long description and then for the short lets do something like
```
Get all repositories for a group
```

src/cmd/group.go
```diff
@@ -38,9 +39,161 @@ EOF
...
+	Long:       `Get descendant repositories for a group`,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID", "ALIAS"},
+	Run: func(cmd *cobra.Command, args []string) {
+		key := args[0]
+		var group *opslevel.Group
+		var err error
+		if common.IsID(key) {
```
rocktavious, line 96: Since this is repeated more then 3 times it should be pull out to a reusable func

src/cmd/group.go
```diff
@@ -38,9 +39,161 @@ EOF
...
+			group, err = getClientGQL().GetGroup(key)
+		} else {
+			group, err = getClientGQL().GetGroupWithAlias(key)
+		}
+		cobra.CheckErr(err)
+		common.PrettyPrint(group)
+		common.WasFound(group.Id == nil, key)
```
rocktavious, line 59: this `WasFound` method needs to come before the `PrettyPrint` so the error is thrown first

rocktavious (changes requested):
Just a few things to fix up and some redundant code that can be DRY'd up.

src/cmd/group.go
eochoa260: I thought the same thing when I checked to see what it did. I'll move it to before and check where I got this from to see if the same needs to be done elsewhere.

src/cmd/group.go
eochoa260: Yeah, and I got it from somewhere else so I'm sure it's the same case for other get commands

src/cmd/group.go
rocktavious: IMO more explicit is always better unless its customer facing.

rocktavious (commented):
LGTM

rocktavious (approved):

PR #28, opened 2022-04-25
Eo/add csv output

eochoa260:


src/cmd/service.go
```diff
@@ -85,6 +85,11 @@ var listServiceCmd = &cobra.Command{
 		cobra.CheckErr(err)
 		if isJsonOutput() {
 			common.JsonPrint(json.MarshalIndent(list, "", "    "))
+		} else if isCsvOutput() {
+			fmt.Printf("%s\n", strings.Join([]string{"NAME", "ID", "ALIASES"}, ","))
```
eochoa260, line 89: Not sure if we need something like NewTabWriter after looking at it more closely since we don't need to handle spacing for csv like we do when printing in table format.  I can add it back in but I noticed that the Join seemed to be working fine in this case.

src/cmd/service.go
```diff
@@ -85,6 +85,11 @@ var listServiceCmd = &cobra.Command{
...
+			for _, item := range list {
+				fmt.Printf("%s,%s,[%s]\n", item.Name, item.Id, strings.Join(item.Aliases, ","))
```
eochoa260, line 91: Add `[]` around aliases since it was hard to differentiate it from NAME and ID fields but might not be necessary.

rocktavious: hmm - is that valid CSV?  I think the extra comma's from `strings.Join(item.Aliases, ",")` will cause problems

rocktavious (commented):
I think it was mainly just around that fact that now the help will say "csv" is an allowed output type but it will only work for `services` listing.

But this is fine - we just won't ship anything with this yet.  We need to move onto on-prem.

There is 1 change/question about valid CSV format syntax.

rocktavious:
@eochoa260 - looks like we just need to join Aliases with an empty space IE `" "`

Here is some examples i pulled online - these have quotes around a cell value that has comma's - i don't like that - lets just do what the `America/Indiana/Indianapolis` has but instead of `/` we use space
```
317-987-6543,America/Indiana/Indianapolis,175,1003,"123 Anystreet, Indianapolis, IN"

src/cmd/service.go
eochoa260: Yeah, I think you're right.  Sounds like we would need to use a different delimiter there.

eochoa260: >Remember that the enclosing character (typically double quotes) must be used when required, such as when the delimiter appears in a field.

rocktavious: ya to be safe i'd use `|` or `/` or any other character you can think of.  We just need to standardize it across the board.

eochoa260: Ok, updated it to use the CSV writer but that defaults to quotes if there's a comma in a field.  Looking at the output, using `/` makes it look a bit like a file path but the `|` blends in a bit more so not sure which would be better.  Went with `/` for now.

PR #32, opened 2022-06-04
Add User commands

rocktavious:
This adds the create, list and delete user commands because the opslevel-go portion is now implemented.

I'm leaving off get user because it requires a user ID and thats not a great UX.  The API will eventually support looking up an user by email so once we do that we will implement that in the CLI.

### DIFF
--- a/src/cmd/team.go
+++ b/src/cmd/team.go
@@ -78,7 +78,7 @@ var createContactCmd = &cobra.Command{
 		case string(opslevel.ContactTypeWeb):
 			contactInput = opslevel.CreateContactWeb(address, displayName)
 		}
-		contact, err := getClientGQL().AddContact(&team.TeamId, contactInput)
+		contact, err := getClientGQL().AddContact(team.TeamId.Alias, contactInput)
 		cobra.CheckErr(err)
 		if contact.Id == nil {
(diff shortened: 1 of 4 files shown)


src/cmd/team.go
```diff
@@ -78,7 +78,7 @@ var createContactCmd = &cobra.Command{
 		case string(opslevel.ContactTypeWeb):
 			contactInput = opslevel.CreateContactWeb(address, displayName)
 		}
-		contact, err := getClientGQL().AddContact(&team.TeamId, contactInput)
+		contact, err := getClientGQL().AddContact(team.TeamId.Alias, contactInput)
```
rocktavious, line 81: Seems like the signature of this function changed - not sure why but I fixed it

src/go.mod
```diff
@@ -56,4 +56,4 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
-//replace github.com/opslevel/opslevel-go => ./submodules/opslevel-go
+replace github.com/opslevel/opslevel-go => ./submodules/opslevel-go
```
rocktavious, line 59: because i'm using the in progress user api stuff

eochoa260 (approved):
LGTM

PR #41, opened 2022-07-21
Eo/apidocs add warning

eochoa260:


src/cmd/document.go
```diff
@@ -34,7 +36,12 @@ opslevel create document my-service -r services -t openapi -i xxxxx -f swagger.j
 			SetHeader("Content-Type", "application/octet-stream").
```
rocktavious, line 36: The correct way to get a custom response object so you don't have to do the marshalling with go-resty is - https://pkg.go.dev/github.com/go-resty/resty/v2#Request.SetResult

I've put in a general purpose response type for this - https://github.com/OpsLevel/opslevel-go/blob/main/clientRest.go#L7
but it looks like we need to add "message" to that

so then the code would then be something like
```
var result opslevel.RestResponse
response, err := getClientRest().R().
			SetBody(fileContents).
			SetHeader("Content-Type", "application/octet-stream").
			SetResult(result).
			Post(integrationURL)
if result.result == "invalid_format" {
...			
```

eochoa260: Ah ok, yeah don't know how I missed that.  I see it in their readme now.  I'll make that change.

rocktavious (approved):
LGTM

PR #44, opened 2022-08-03
Adding function to read files in policies

eochoa260:
Opening PR to try to get it reviewed as I go since I'll be OOO tomorrow.

### DIFF
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -7,6 +7,7 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/types"
+	"github.com/rs/zerolog/log"
 	cobra "github.com/spf13/cobra"
 	"io/ioutil"
 	"os"
@@ -64,9 +65,13 @@ opslevel run policy -f policy.rego | jq
 
 func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	if str, ok := a.Value.(ast.String); ok {
-		contents, err := os.ReadFile(string(str))
-		cobra.CheckErr(err)
-		return ast.StringTerm(string(contents)), nil
+		if _, err := os.Stat(string(str)); err != nil {
+			log.Warn().Msgf("%s", err)
+		} else {
+			contents, err := os.ReadFile(string(str))
+			cobra.CheckErr(err)
+			return ast.StringTerm(string(contents)), nil
+		}
 	}
 	return nil, nil
 }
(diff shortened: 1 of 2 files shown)


src/cmd/policy.go
```diff
@@ -54,6 +62,15 @@ opslevel run policy -f policy.rego | jq
 	},
 }
 
+func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
+	if str, ok := a.Value.(ast.String); ok {
+		contents, err := os.ReadFile(string(str))
```
eochoa260, line 67: I know there's different ways to read a file, e.g. line-by-line, chunks, etc.  My understanding is chunks would be better in the case of larger files so I can update this if we'd prefer to read it that way.  I can focus on that after lunch if so.

rocktavious: Chunking is best if you are going to be reading through the file and process data as it comes in reusing the buffer as you move through the data.  Unfortunately because we want to load up the contents of the file fully and then have Rego run more policy stuff on the content we need to fully load the contents into more - so this is the best way.

rocktavious, line 67: do we need to do filepath exists validation here?

eochoa260: Ok, that was one of my concerns. I wasn't sure how the content would look on the Rego side.

rocktavious (commented):
1 question

src/cmd/policy.go
rocktavious: Ya i'm not entirely sure here - whats the user experience like when you use a policy with a file thats not present?

eochoa260: It exits with the error so it doesn't read the other files.  Changed it to warn if filepath doesn't exist.

eochoa260:
@rocktavious I think this is ready.  I was able to read files and check for text with a policy to check for some keywords.

src/cmd/policy.go
```diff
@@ -54,6 +64,26 @@ opslevel run policy -f policy.rego | jq
...
 }
 
+func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
+	if str, ok := a.Value.(ast.String); ok {
+		if _, err := os.Stat(string(str)); err != nil {
+			log.Warn().Msgf("%s", err)
+		} else {
+			file, err := os.Open(string(str))
+			cobra.CheckErr(err)
+			defer file.Close()
```
rocktavious, line 74: do this before the cobra check error exits

eochoa260: ah yeah, good catch

rocktavious (approved):
LGTM

PR #45, opened 2022-08-10
Eo/add GitHub rego function

eochoa260:
```
$ cat policy.rego                                          
package opslevel

import future.keywords

description := opslevel.repo.github("OpsLevel", "opslevel-go")["description"]
language := opslevel.repo.github("OpsLevel", "opslevel-go")["language"]                                                                                                                                                                                                                                                  

$ go run main.go run policy -f policy.rego --github-token <github-token>
{"description":"A Go client for interacting with OpsLevel","language":"Go"}
```

### DIFF
--- a/src/cmd/document.go
+++ b/src/cmd/document.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/opslevel/opslevel-go/v2022"
 	"github.com/rs/zerolog/log"
-	cobra "github.com/spf13/cobra"
+	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
--- a/src/cmd/document.go
+++ b/src/cmd/document.go
@@ -7,7 +7,7 @@ import (
 
 	"github.com/opslevel/opslevel-go/v2022"
 	"github.com/rs/zerolog/log"
-	"github.com/spf13/cobra"
+	cobra "github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -102,8 +102,13 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 		return nil, err
 	}
 
-	if org == "" || repo == "" {
-		log.Error().Msgf("Please provide a valid org and repo")
+	if org == "" {
+		log.Error().Msgf("Please provide a valid org")
+		return nil, nil
+	}
+
+	if repo == "" {
+		log.Error().Msgf("Please provide a valid repo")
 		return nil, nil
 	}
 
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -137,4 +137,8 @@ func init() {
 	runCmd.AddCommand(policyCmd)
 
 	policyCmd.Flags().StringP("file", "f", "-", "File to read Rego policy from. Defaults to reading from stdin.")
+	policyCmd.PersistentFlags().String("github-token", "", "The Github API token to use when calling opslevel.repo.github function within a Rego policy. Overrides environment variable 'GITHUB_API_TOKEN'")
+
+	viper.BindPFlags(policyCmd.PersistentFlags())
+	viper.BindEnv("github-token", "GITHUB_API_TOKEN")
 }
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -98,7 +98,8 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 	var org, repo string
 	if err := ast.As(a.Value, &org); err != nil {
 		return nil, err
-	} else if ast.As(b.Value, &repo); err != nil {
+	}
+	if err := ast.As(b.Value, &repo); err != nil {
 		return nil, err
 	}
 
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -13,10 +13,17 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type regoInput struct {
 	Files []string `json:"files"`
+}
+
+type GithubRepo struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Language    string `json:"language"`
 }
 
 var policyCmd = &cobra.Command{
@@ -109,10 +116,16 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 		SetHeader("Accept", "application/vnd.github+json").
 		SetHeader("Authorization", fmt.Sprintf("token %v", githubToken)).
 		Get(fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo))
-
 	cobra.CheckErr(err)
 
-	return ast.StringTerm(response.String()), nil
+	if response.IsError() == true {
+		log.Error().Msgf("%d: %s", response.StatusCode(), response)
+		return nil, err
+	}
+
+	reader := strings.NewReader(response.String())
+	v, err := ast.ValueFromReader(reader)
+	return ast.NewTerm(v), nil
 }
 
 func init() {
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -100,11 +100,15 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 	} else if ast.As(b.Value, &repo); err != nil {
 		return nil, err
 	}
-	fmt.Println(org)
-	fmt.Println(repo)
+
+	flags := runCmd.Flags()
+	githubToken, err := flags.GetString("github-token")
+	cobra.CheckErr(err)
 
 	response, err := getClientRest().R().
-		Get("https://httpbin.org/get")
+		SetHeader("Accept", "application/vnd.github+json").
+		SetHeader("Authorization", fmt.Sprintf("token %v", githubToken)).
+		Get(fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo))
 
 	cobra.CheckErr(err)
 
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -110,6 +110,9 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 
 	flags := runCmd.Flags()
 	githubToken, err := flags.GetString("github-token")
+	if githubToken == "" {
+		githubToken = os.Getenv("GITHUB_API_TOKEN")
+	}
 	cobra.CheckErr(err)
 
 	response, err := getClientRest().R().
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -117,7 +117,7 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 
 	response, err := getClientRest().R().
 		SetHeader("Accept", "application/vnd.github+json").
-		SetHeader("Authorization", fmt.Sprintf("token %v", githubToken)).
+		SetHeader("Authorization", fmt.Sprintf("token %s", githubToken)).
 		Get(fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo))
 	cobra.CheckErr(err)
 
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -103,12 +103,12 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 	}
 
 	if org == "" {
-		log.Error().Msgf("Please provide a valid org")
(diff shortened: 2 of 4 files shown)


src/cmd/run.go
```diff
@@ -12,4 +10,5 @@ var runCmd = &cobra.Command{
 
 func init() {
 	rootCmd.AddCommand(runCmd)
+	runCmd.PersistentFlags().String("github-token", "", "Github API Token. Overrides environment variable 'GITHUB_API_TOKEN'")
```
eochoa260, line 13: Still working on getting this to work with env var too but doesn't seem to work when I add the viper bindenv like in root.go.

src/cmd/policy.go
```diff
@@ -84,6 +91,30 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+	cobra.CheckErr(err)
+
+	response, err := getClientRest().R().
+		SetHeader("Accept", "application/vnd.github+json").
+		SetHeader("Authorization", fmt.Sprintf("token %v", githubToken)).
+		Get(fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo))
+
+	cobra.CheckErr(err)
+
+	return ast.StringTerm(response.String()), nil
```
eochoa260, line 115: This still needs work... need to find a way to return an object instead of a string.

src/cmd/policy.go
```diff
@@ -84,6 +98,39 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+	flags := runCmd.Flags()
+	githubToken, err := flags.GetString("github-token")
+	if githubToken == "" {
+		githubToken = os.Getenv("GITHUB_API_TOKEN")
+	}
+	cobra.CheckErr(err)
+
+	response, err := getClientRest().R().
+		SetHeader("Accept", "application/vnd.github+json").
+		SetHeader("Authorization", fmt.Sprintf("token %s", githubToken)).
```
rocktavious, line 120: for readability (and since the line is already very long) i would move the 2 `fmt.Sprintf` out to variables and just pump the variables in here.

rocktavious (changes requested):
A bunch of questions and a few changes

src/cmd/document.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/opslevel/opslevel-go/v2022"
 	"github.com/rs/zerolog/log"
-	"github.com/spf13/cobra"
+	cobra "github.com/spf13/cobra"
```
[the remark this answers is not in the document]
eochoa260: This was really annoying and I probably should take a look in Goland to figure out what's going on but it kept removing this import and saying that cobra was undefined.  So I added cobra in there and it's keeping it now.

src/cmd/policy.go
```diff
@@ -13,12 +13,19 @@ import (
...
 	"os"
 	"path/filepath"
+	"strings"
 )
 
 type regoInput struct {
 	Files []string `json:"files"`
 }
 
+type GithubRepo struct {
```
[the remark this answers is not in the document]
eochoa260: Forgot to delete this.  I added this when I was trying to set result to a struct.  Don't need to do it now since I converted the response body to io.Reader and did the same thing that the OPA docs had.

src/cmd/policy.go
```diff
@@ -84,6 +98,39 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	return nil, nil
 }
 
+func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, error) {
+
+	var org, repo string
+
+	if err := ast.As(a.Value, &org); err != nil {
```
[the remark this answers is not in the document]
eochoa260: My understanding is that if `a` is nil then `a.Value` would error but I'll look into this because I'm not sure

src/cmd/policy.go
```diff
@@ -84,6 +98,39 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+		return nil, err
+	} else if ast.As(b.Value, &repo); err != nil {
+		return nil, err
+	}
+
+	flags := runCmd.Flags()
```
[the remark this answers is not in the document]
eochoa260: Sweet! I'll add that in.  Thanks!

src/cmd/policy.go
eochoa260: That's a good point.  I'll do that.

src/cmd/policy.go
```diff
@@ -84,6 +98,39 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+		SetHeader("Authorization", fmt.Sprintf("token %s", githubToken)).
+		Get(fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo))
+	cobra.CheckErr(err)
+
+	if response.IsError() == true {
+		log.Error().Msgf("%d: %s", response.StatusCode(), response)
+		return nil, err
+	}
+
+	reader := strings.NewReader(response.String())
```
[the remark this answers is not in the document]
eochoa260: Yep, this is working.  Didn't need a struct after all.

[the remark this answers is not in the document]
rocktavious: so in a rego policy you can do this?

```
github_metadata := opslevel.repo.github("org", "repo")
language := github_metadata.language
```

src/cmd/policy.go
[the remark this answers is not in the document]
rocktavious: hmm - ya this needs a test - its possible the rego machinery doesn't allow `a` to be nil because of the schema in but if it doesn't we'd get a panic on `a.Value` because accessing a nil struct field.

src/cmd/document.go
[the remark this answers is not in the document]
rocktavious: ya might need to do a `go mod download` locally and/or restart goland

src/cmd/policy.go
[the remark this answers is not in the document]
rocktavious: sorry i didn't see your example at the top.  While that works i'm concerned it makes our UX worse then if we had a struct.

But if we have a struct then it makes it hard to get at new / non-implemented things.

I'm fine with letting us go this route for now.  But i'm fear it will cause long term policy writing pain as just do blind access into arbitrary json - IE
```
github_metadata := opslevel.repo.github("org", "repo")
language := github_metadata["language"]    <-- this part right here could fail with "key not found"
```

src/cmd/policy.go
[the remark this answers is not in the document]
eochoa260: tried these 2 tests and this is what I got, respectively:
`opslevel.repo.github()`
`opslevel.repo.github("","")`
```
$ go run main.go run policy -f github-repo.rego
Error: 1 error occurred: test.rego:9: rego_type_error: opslevel.repo.github: arity mismatch
	have: (???)
	want: (string, string)
exit status 1
$ go run main.go run policy -f github-repo.rego
5:26PM ERR 404: {"message":"Not Found","documentation_url":"https://docs.github.com/rest"}
```

src/cmd/policy.go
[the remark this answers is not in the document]
eochoa260: That's a very good point, I figured it would be nice to be able to print what's returned in the policy and pick and choose which fields we want but I agree that once we have a pattern it will be better to have a struct defined to only pull what we need.  Would that be a pain if customers want to build custom policies and we don't have the fields they're looking for?

src/cmd/policy.go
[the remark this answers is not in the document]
eochoa260: I'll have it error when it's an empty string just so that it doesn't make it to the github api call.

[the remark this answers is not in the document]
eochoa260: Similar to what we did in the read file function, I'm assuming we'd want to log a message and exit the function call but continue the run execution in case there are other function calls that are correctly called?  Not sure how the app will expect bad function calls in a rego policy.

src/cmd/policy.go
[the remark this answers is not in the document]
rocktavious: Ya the struct would limit our users - but also protect them.  So thats why i'm torn.

Lets go with it for now and we can add in the struct later.

src/cmd/policy.go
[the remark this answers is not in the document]
rocktavious: The logs emitted by the CLI will probably be discoverable somehow - so whatever debugging information we can emit in the logs the better.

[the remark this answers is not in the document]
eochoa260: Sounds good, just had it return error message, exit the function but proceed with the run.

src/cmd/policy.go
```diff
@@ -96,13 +96,17 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
 	var org, repo string
-
 	if err := ast.As(a.Value, &org); err != nil {
 		return nil, err
 	} else if ast.As(b.Value, &repo); err != nil {
 		return nil, err
 	}
 
+	if org == "" || repo == "" {
+		log.Error().Msgf("Please provide a valid org and repo")
```
rocktavious, line 106: this split this into 2 and give a message to explicitly call out which one

src/cmd/policy.go
```diff
@@ -102,8 +102,13 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
 		return nil, err
 	}
 
-	if org == "" || repo == "" {
-		log.Error().Msgf("Please provide a valid org and repo")
+	if org == "" {
+		log.Error().Msgf("Please provide a valid org")
```
rocktavious, line 106: is the customer going to know this is for "github" ?  I think just logging this they probably won't as it won't attribute to the policy line.

Maybe we should return this an error?

eochoa260: Very true, didn't think about that.  Should I add some generic prefix to the error messages?  e.g. `opslevel.repo.github(<org>, <repo>) failed: Please provide a valid org?` Not sure how explicit we want to get?

rocktavious: that or see if rego does this by changing it to be

```
if org == "" {
	return nil, fmt.Errorf("Please provide a valid value for github org")
}
```

eochoa260: I tried that earlier and it doesn't output it for some reason.  Probably something I would need to change in the eval?  I can take a look though.

rocktavious: ok well if there is nothing built in - we just need to make sure we give as much context as possible to help debug the issue.  Terse error messages are bad and they are even worse in a black box system like we are building with "repo analysis"

src/cmd/policy.go
```diff
@@ -84,6 +93,45 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+	githubAPIUrl := fmt.Sprintf("https://api.github.com/repos/%v/%v", org, repo)
+
+	response, err := getClientRest().R().
+		SetHeader("Accept", "application/vnd.github+json").
+		SetHeader("Authorization", authorizationHeader).
+		Get(githubAPIUrl)
+	cobra.CheckErr(err)
+
+	if response.IsError() == true {
+		log.Error().Msgf("%d: %s", response.StatusCode(), response)
```
rocktavious, line 126: while we are on error messages - we could improve this one to call out it was an api call to github and this is the API response.

maybe something like

```
log.Error().Msgf("error requesting github repo metadata. CODE: %d REASON: %s", response.StatusCode(), response)
```

src/cmd/policy.go
```diff
@@ -84,6 +93,45 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 	return nil, nil
 }
 
+func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, error) {
+
+	var org, repo string
+	if err := ast.As(a.Value, &org); err != nil {
+		return nil, err
+	} else if ast.As(b.Value, &repo); err != nil {
```
rocktavious, line 101: this shouldn't be an else?

eochoa260: 🤷 That's how they have it in their example.  I think it works as an `else if` since they're just checking for errors when they store the value in each var but I could also see it working as just a separate if block.

src/cmd/policy.go
eochoa260: For sure, I'll add the function call that way it can be at least be identified in the policy.

src/cmd/policy.go
rocktavious: ya lets just make it 2 if as thats confusing if it really does work.  As it reads like the 2nd if will get skipped

src/cmd/policy.go
```diff
@@ -84,6 +93,45 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
...
+		return nil, err
+	}
+
+	if org == "" {
+		log.Error().Msgf("opslevel.repo.github(\"%s\", \"%s\") failed: Please provide a valid org", org, repo)
```
rocktavious, line 106: ❤️

src/cmd/policy.go
eochoa260: Done :-)

rocktavious (approved):
LGTM

PR #46, opened 2022-08-19
Add function to check maturity level of a service

eochoa260:
Example:
```
$ cat maturity-check.rego 
package opslevel

import future.keywords

default can_deploy_dev := false
default can_deploy_staging := false

alias := opslevel.service_maturity_level("cert-manager")["Alias"]
index := opslevel.service_maturity_level("cert-manager")["Index"]

can_deploy_dev {
    index >= 0
}

can_deploy_staging {
    alias == "silver"
}

$ go run main.go run policy -f maturity-check.rego
{"alias":"beginner","can_deploy_dev":true,"can_deploy_staging":false,"index":0}
```

### DIFF
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -62,10 +62,10 @@ opslevel run policy -f policy.rego | jq
 					Memoize: true,
 				},
 				RegoFuncGetGithubRepo),
-			rego.Function2(
+			rego.Function1(
 				&rego.Function{
-					Name: "opslevel.service_maturity_is",
-					Decl: types.NewFunction(types.Args(types.S, types.S), types.B),
+					Name: "opslevel.service_maturity_level",
+					Decl: types.NewFunction(types.Args(types.S), types.A),
 				},
 				RegoFuncGetMaturity),
 			rego.Input(input),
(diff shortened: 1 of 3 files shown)


src/cmd/policy.go
```diff
@@ -133,6 +139,37 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+	if level == "" {
+		log.Error().Msgf("opslevel.service_maturity_is(\"%s\", \"%s\") failed: Please provide a valid repo", alias, level)
+		return nil, nil
+	}
+
+	client := getClientGQL()
+	service, err := client.GetServiceMaturityWithAlias(alias)
+	cobra.CheckErr(err)
+	serviceLevel := service.MaturityReport.OverallLevel.Alias
+	if serviceLevel == level {
```
rocktavious, line 166: So we have to do a more complex check here.  The way i see this being used is to ensure the service level is `>= "silver"`. Maybe that means we also need to change the name of the function too? 🤷 

So we may need to use the level index or something and do a >= test.

rocktavious (changes requested):
Let me know if that doesn't make sense - and we can sync on Monday

src/cmd/policy.go
eochoa260: Ah ok, that's a good point.  Would it be better to change it to just return the level and let the policy writer add whatever logic they want in their rego policy to gate deployments?

rocktavious: @eochoa260 ouu thats a great idea - that means is more like "get service level index" so they can choose what kind of logic to run on it `>=` or `<=` or even just `==`. ❤️

eochoa260: Ok, updated it to return the OverallLevel for a service as an object so users can get the Name, Index, Alias, etc and run any comparison they want in the Rego policy. 😃

rocktavious (approved):
LGTM

PR #48, opened 2022-08-23
Add Rego function to get gitlab project metadata

eochoa260:
```
$ cat gitlab-repo.rego 
package opslevel

import future.keywords

contains_value(input_list, value) {
    contains(input_list[_], value)
}

repo_name := opslevel.repo.gitlab("jklabsinc/OpsLevel")["name"]
repo_description := opslevel.repo.gitlab("jklabsinc/OpsLevel")["description"]
repo_languages := opslevel.repo.gitlab("jklabsinc/OpsLevel")["languages"]

$ go run main.go run policy -f gitlab-repo.rego                   
{"repo_description":"Source for app.opslevel.com.   Level up your DevOps teams.","repo_languages":{"CSS":1.91,"HTML+ERB":0.41,"JavaScript":26.03,"Ruby":59.17,"Vue":11.95},"repo_name":"OpsLevel"}
```

### DIFF
--- a/src/cmd/policy.go
+++ b/src/cmd/policy.go
@@ -49,22 +49,22 @@ type astMarshallable interface {
 	commonRepoMetadata | opslevel.Level
 }
 
-func githubResponseToRepoMetadata(response githubResponse) commonRepoMetadata {
-	metadata := commonRepoMetadata{}
-	metadata.Name = response.Name
-	metadata.Description = response.Description
-	metadata.Language = response.Language
-	metadata.Languages = response.Languages
-	return metadata
+func (r *githubResponse) toRepoMetadata() commonRepoMetadata {
+	return commonRepoMetadata{
+		Name:        r.Name,
+		Description: r.Description,
+		Language:    r.Language,
+		Languages:   r.Languages,
+	}
 }
 
-func gitlabResponseToRepoMetadata(response gitlabResponse) commonRepoMetadata {
-	metadata := commonRepoMetadata{}
-	metadata.Name = response.Name
-	metadata.Description = response.Description
-	metadata.Language = response.Language
-	metadata.Languages = response.Languages
-	return metadata
+func (r *gitlabResponse) toRepoMetadata() commonRepoMetadata {
+	return commonRepoMetadata{
+		Name:        r.Name,
+		Description: r.Description,
+		Language:    r.Language,
+		Languages:   r.Languages,
+	}
 }
 
 func toASTValue[T astMarshallable](input T) (ast.Value, error) {
@@ -178,6 +178,7 @@ func RegoFuncReadFile(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error) {
 			file, err := os.Open(string(str))
 			defer file.Close()
 			if err != nil {
+				log.Error().Err(err).Msg("")
 				return nil, err
 			}
 
@@ -196,9 +197,11 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 
 	var org, repo string
 	if err := ast.As(a.Value, &org); err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 	if err := ast.As(b.Value, &repo); err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -226,6 +229,7 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 		SetResult(&result).
 		Get(githubAPIUrl)
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -241,6 +245,7 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 		SetResult(&result.Languages).
 		Get(githubAPIUrl + "/languages")
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -251,10 +256,11 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term, 
 	}
 
 	result.Languages = convertToPercentage(result.Languages)
-	repoMetadata := githubResponseToRepoMetadata(result)
+	repoMetadata := result.toRepoMetadata()
 
 	v, err := toASTValue(repoMetadata)
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 	return ast.NewTerm(v), nil
@@ -264,6 +270,7 @@ func RegoFuncGetGitlabRepo(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, err
 
 	var path string
 	if err := ast.As(a.Value, &path); err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -284,6 +291,7 @@ func RegoFuncGetGitlabRepo(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, err
 		SetResult(&result).
 		Get(gitlabAPIUrl)
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -298,6 +306,7 @@ func RegoFuncGetGitlabRepo(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, err
 		SetResult(&result.Languages).
 		Get(gitlabAPIUrl + "/languages")
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
@@ -308,9 +317,10 @@ func RegoFuncGetGitlabRepo(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, err
 	}
 
 	result.Language = getKeyFromMapByMaxValue(result.Languages)
-	repoMetadata := gitlabResponseToRepoMetadata(result)
+	repoMetadata := result.toRepoMetadata()
 	v, err := toASTValue(repoMetadata)
 	if err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 	return ast.NewTerm(v), nil
@@ -320,6 +330,7 @@ func RegoFuncGetMaturity(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error
 
 	var alias string
 	if err := ast.As(a.Value, &alias); err != nil {
+		log.Error().Err(err).Msg("")
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


src/cmd/policy.go
```diff
@@ -136,6 +144,61 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+	if path == "" {
+		log.Error().Msgf("opslevel.repo.gitlab(\"%s\") failed: Please provide a valid Gitlab project path", path)
+		return nil, nil
+	}
+
+	escapedPath := url.QueryEscape(path)
+	gitlabToken := viper.GetString("gitlab-token")
+	gitlabAPIUrl := fmt.Sprintf("https://gitlab.com/api/v4/projects/%s", escapedPath)
+
+	type GitlabRepoMetadata struct {
```
eochoa260, line 167: Went with a struct this time since I had to make 2 calls this time and thought it might be better to explicitly build up an object in this case? Not sure if that's the right approach or if I should just add languages result to the first response.

rocktavious: Ok - this is a good test to see which route works better long term.

should the struct be defined outside the function?
should the struct be private (first letter lowercase)?
How do we intend to document the available fields?

src/cmd/policy.go
```diff
@@ -136,6 +144,61 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+		Name        string             `json:"name,omitempty"`
+		Description string             `json:"description,omitempty"`
+		Languages   map[string]float32 `json:"languages,omitempty"`
```
rocktavious, line 170: is the data format of this different from the Github one?  We should keep them insync in terms of data format exposed to the policy.

src/cmd/policy.go
```diff
@@ -136,6 +144,61 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+
+	if languagesResponse.IsError() == true {
+		log.Error().Msgf("error requesting Gitlab repo languages. CODE: %d: REASON: %s", response.StatusCode(), response)
+		return nil, nil
+	}
+
+	repoMetadataMarshal, err := json.Marshal(repoMetadata)
+	cobra.CheckErr(err)
+	reader := strings.NewReader(string(repoMetadataMarshal))
+	v, err := ast.ValueFromReader(reader)
```
rocktavious, line 200: is this the 3rd time we've done this kind of marshaling to AST?  Should be build a reusable function that bundles all this up and takes in an interface (or even better Generics)?

rocktavious (changes requested):
Overall the logic is good.  I have a few questions and 1 that i think will result in changes.

src/cmd/policy.go
eochoa260: Yeah, this should be defined outside of the function and should be private for now unless we need to end up using it elsewhere.  I imagine it'll have to be documented when we document each function that we've added to Rego?

src/cmd/policy.go
eochoa260: Yeah, we're just getting top language on the Github side and Gitlab doesn't return that so I made the extra call to get Languages which is similar to what Github would do.  Not sure what the Github one returns for each language though, lines? Either way it uses these numbers to calculate a percentage which I assume we would want to do when we start pulling in languages?  Maybe I should add a Language field similar to what Github and set it equal to the highest percentage?
```
{
  "Go": 104748,
  "Shell": 1335,
  "Dockerfile": 379
}
```

src/cmd/policy.go
rocktavious: Ok

src/cmd/policy.go
rocktavious: We don't have do this for this ticket but we might want to enqueue a backlog ticket for documenting our custom functions.

src/cmd/policy.go
eochoa260: Just so that I'm on the same page, are you saying that we should have the same fields returned for each function?  Or as close as possible?  Right now Github is returning everything so I agree that it's very different UX for each function as it is.  Should I go ahead and make this struct repo metadata and use it for both the github and gitlab functions? That way it returns the same info?

eochoa260:
Output from github and gitlab functions after recent changes:
```
$ go run main.go run policy -f github-repo.rego                    
{"description":"Opslevel Command Line Utility","language":"Go","languages":{"Dockerfile":0.4,"Go":98.4,"Shell":1.2},"name":"cli"}

$ go run main.go run policy -f gitlab-repo.rego 
{"description":"Source for app.opslevel.com.   Level up your DevOps teams.","language":"Ruby","languages":{"CSS":1.91,"HTML+ERB":0.41,"JavaScript":25.99,"Ruby":59.24,"Vue":11.93},"name":"OpsLevel"}
```

src/cmd/policy.go
```diff
@@ -21,6 +24,63 @@ type regoInput struct {
...
+	Description string             `json:"description,omitempty"`
+	Language    string             `json:"language,omitempty"`
+	Languages   map[string]float64 `json:"languages,omitempty"`
+}
+
+type genericStruct interface {
+	gitlabRepoMetadata | githubRepoMetadata | opslevel.Level
+}
+
+func structToValue[S genericStruct](s S) ast.Value {
```
eochoa260, line 45: I don't know if I did this right but I thought it was really cool either way. lol

src/cmd/policy.go
```diff
@@ -21,6 +24,63 @@ type regoInput struct {
...
+	marshalData, err := json.Marshal(s)
+	cobra.CheckErr(err)
+	reader := strings.NewReader(string(marshalData))
+	v, err := ast.ValueFromReader(reader)
+	cobra.CheckErr(err)
+	return v
+}
+
+func getMaxFromMap(m map[string]float64) string {
```
eochoa260, line 54: Feels weird having to write this function here since I could see it being a common function but left it here for now.

src/cmd/policy.go
```diff
@@ -21,6 +24,63 @@ type regoInput struct {
...
+		if v > max {
+			max = v
+			output = k
+		}
+	}
+	return output
+}
+
+func convertToPercentage(m map[string]float64) map[string]float64 {
+	total := float64(0)
```
eochoa260, line 67: Same as above

src/cmd/policy.go
```diff
@@ -21,6 +24,63 @@ type regoInput struct {
...
+		total += v
+	}
+
+	for k, v := range m {
+		output[k] = roundFloat((v/total)*100, 1)
+	}
+	return output
+}
+
+func roundFloat(val float64, precision uint) float64 {
```
eochoa260, line 79: Found this with a quick search for getting sig figs down to 1 to match what it shows in the UI.  Not sure if this is necessary.

src/cmd/policy.go
```diff
@@ -21,6 +24,63 @@ type regoInput struct {
...
+}
+
+type githubRepoMetadata struct {
+	Name        string             `json:"name,omitempty"`
+	Description string             `json:"description,omitempty"`
+	Language    string             `json:"language,omitempty"`
+	Languages   map[string]float64 `json:"languages,omitempty"`
+}
+
+type genericStruct interface {
```
rocktavious, line 41: i would give this a better name - maybe something related to AST marshaling - maybe `astMarshallable` or something

src/cmd/policy.go
rocktavious: ya eventually we could pull this out to opslevel-common - but fine for now

rocktavious: a better function name would be `getKeyFromMapByMaxValue`

its more verbose but it conveys exactly what the function does - self documenting

src/cmd/policy.go
rocktavious: wow surprised this isn't built into the standard lib

rocktavious (changes requested):
Alright i know thats alot let me know if you want to sync over zoom on any of it.

src/cmd/policy.go
eochoa260: Ok, awesome! Making those changes now.

src/cmd/policy.go
```diff
@@ -21,6 +24,87 @@ type regoInput struct {
...
+	Description string             `json:"description,omitempty"`
+	Language    string             `json:"language,omitempty"`
+	Languages   map[string]float64 `json:"languages,omitempty"`
+}
+
+type astMarshallable interface {
+	commonRepoMetadata | opslevel.Level
+}
+
+func githubResponseToRepoMetadata(response githubResponse) commonRepoMetadata {
```
eochoa260, line 52: Tried to use a generic function here but it looks like you can't access fields if you use structs as the generic type 😞 
https://stackoverflow.com/questions/71530811/how-to-access-shared-field-in-structs-in-generics-in-go-1-18-i-get-error-type

src/cmd/policy.go
```diff
@@ -110,32 +203,116 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+	gitlabToken := viper.GetString("gitlab-token")
+	gitlabAPIUrl := fmt.Sprintf("https://gitlab.com/api/v4/projects/%s", escapedPath)
+
+	var result gitlabResponse
+
+	response, err := getClientRest().R().
+		SetHeader("PRIVATE-TOKEN", gitlabToken).
+		SetResult(&result).
+		Get(gitlabAPIUrl)
+	if err != nil {
```
eochoa260, line 286: Should I be logging the error here similar to line 292?

rocktavious: yes probably

src/cmd/policy.go
```diff
@@ -110,32 +203,116 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+		log.Error().Err(err).Msgf("")
+		return nil, err
+	}
+
+	languagesResponse, err := getClientRest().R().
+		SetHeader("PRIVATE-TOKEN", gitlabToken).
+		SetResult(&result.Languages).
+		Get(gitlabAPIUrl + "/languages")
+	if err != nil {
+		return nil, err
```
rocktavious, line 301: should probably log the error here too

src/cmd/policy.go
```diff
@@ -110,32 +203,116 @@ func RegoFuncGetGithubRepo(ctx rego.BuiltinContext, a, b *ast.Term) (*ast.Term,
...
+		err := fmt.Errorf("error requesting Gitlab repo languages. CODE: %d: REASON: %s", response.StatusCode(), response)
+		log.Error().Err(err).Msgf("")
+		return nil, err
+	}
+
+	result.Language = getKeyFromMapByMaxValue(result.Languages)
+	repoMetadata := gitlabResponseToRepoMetadata(result)
+	v, err := toASTValue(repoMetadata)
+	if err != nil {
+		return nil, err
```
rocktavious, line 314: same - log the error

src/cmd/policy.go
```diff
@@ -147,18 +324,21 @@ func RegoFuncGetMaturity(ctx rego.BuiltinContext, a *ast.Term) (*ast.Term, error
...
 	client := getClientGQL()
 	service, err := client.GetServiceMaturityWithAlias(alias)
-	cobra.CheckErr(err)
-	serviceLevel, err := json.Marshal(service.MaturityReport.OverallLevel)
-	cobra.CheckErr(err)
-	reader := strings.NewReader(string(serviceLevel))
-	v, err := ast.ValueFromReader(reader)
-	cobra.CheckErr(err)
+	if err != nil {
+		return nil, err
```
rocktavious, line 335: same - log the error

rocktavious (approved):
Overall the changes are good to merge - just a few nits that could improve it but only if you think they are actually necessary.  Error handling is still something i'm no 100% sure on in terms of how it works in Rego.

src/cmd/policy.go
eochoa260: Oh yeah, that is much better.  I'll make those changes now.  You would think that would be my first thought after how much golang code I've read through recently.  lol

eochoa260:
Thanks for the review @rocktavious! Learned a lot on this one.

PR #54, opened 2022-09-08
Allow tags to be deleted by ID

SegFaultAX:
This simple patch allows the `opslevel delete service tag` command to accept a tag id in addition to a tag name.

### DIFF
--- a/src/cmd/service.go
+++ b/src/cmd/service.go
@@ -228,11 +228,18 @@ var deleteServiceTagCmd = &cobra.Command{
 		if result.Id == nil {
 			cobra.CheckErr(fmt.Errorf("service '%s' not found", serviceKey))
 		}
-		for _, tag := range result.Tags.Nodes {
-			if tagKey == tag.Key || tagKey == tag.Id {
-				getClientGQL().DeleteTag(tag.Id)
-				fmt.Println("Deleted Tag")
-				common.PrettyPrint(tag)
+
+		if common.IsID(tagKey) {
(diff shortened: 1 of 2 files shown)


src/cmd/service.go
```diff
@@ -228,7 +229,7 @@ var deleteServiceTagCmd = &cobra.Command{
 			cobra.CheckErr(fmt.Errorf("service '%s' not found", serviceKey))
 		}
 		for _, tag := range result.Tags.Nodes {
-			if tagKey == tag.Key {
+			if tagKey == tag.Key || tagKey == tag.Id {
```
rocktavious, line 232: This works; but causes unnecessary looping.  We have a function to determine if the passed in string is an ID so it would eliminate the need for looping over all the tags in the case of the user passing in an ID instead of a tag key name.

For example:
```go
		if common.IsID(tagKey) {
			err := getClientGQL().DeleteTag(tagKey)
			cobra.CheckErr(err)
			fmt.Println("Deleted Tag")
		} else {
			for _, tag := range result.Tags.Nodes {
				if tagKey == tag.Key {
```

I also threw in some error handling since it seems we were not doing that properly here.

rocktavious (changes requested):
Thanks for adding this functionality.  There is an alternative way that would eliminate looping over all the tags of the service.

src/cmd/service.go
SegFaultAX: Thanks, I'll update my PR to reflect this change now.

rocktavious (approved):
LGTM

PR #60, opened 2022-09-22
Add 2 features needed for “Magic Catalog”

rocktavious:
This add 2 new flags to the `opslevel run policy` command

`-i ./input.json` - which allows for passing in extra JSON input for use in the policy at `input.data`

`-o ./output.json` - which allows for directly writing all policy evaluation data to a file so it doesn't get intermixed with other output from loggers

A full example would be:

```
cat << EOF > policy.rego
package opslevel
import future.keywords

value := input.data.foo.Can.nest
EOF

cat << EOF > input.json
{"foo": {"Can": {"nest": true}}}
EOF

opslevel run policy -f policy.rego -i input.json -o output.json 

cat output.json
{"value":true}
```


eochoa260 (approved):
LGTM

PR #72, opened 2022-11-09
add CRUD support for team tags

rauhut:


src/cmd/team.go
```diff
@@ -231,14 +361,18 @@ EOF
 
 func init() {
 	createCmd.AddCommand(createTeamCmd)
-	createCmd.AddCommand(createMemberCmd)
-	createCmd.AddCommand(createContactCmd)
+	createTeamCmd.AddCommand(createMemberCmd)
```
rauhut, line 364: I changed some of these since I think they were incorrect and causing an issue. For example, if I would try to create a team member doing something like `create team member Z2lkOi8vb3BzbGV2ZWwvVGVhbS8xMTQ 
 "jackie@opslevel.com"` I would get this error
```
Error: accepts 1 arg(s), received 3
Usage:
  opslevel create team NAME [flags]
```
It thinks I'm trying to create a team; same issue for creating team contacts and team tags. I noticed in `src/cmd/service.go` that we called `AddCommand` on the original create/get/delete command for the object instead of on `createCmd `, and when I updated this file to follow a similar pattern it fixed this issue.

Let me know if this is incorrect and I can revert this change

rauhut: discussed this, and turned out to just be a misunderstanding about how to run this command

src/cmd/team.go
```diff
@@ -88,6 +88,55 @@ var createContactCmd = &cobra.Command{
...
+		if tagAssign {
+			input := opslevel.TagAssignInput{
+				Tags: []opslevel.TagInput{
+					{Key: tagKey, Value: tagValue},
+				},
+			}
+			if common.IsID(teamKey) {
+				input.Id = teamKey
+			} else {
+				input.Alias = teamKey
```
rocktavious, line 118: If we run through this usecase we won't properly assign the tag because we need to set the `TaggableResource` type

src/cmd/team.go
```diff
@@ -88,6 +88,55 @@ var createContactCmd = &cobra.Command{
...
+			result, err = getClientGQL().AssignTags(input)
+		} else {
+			input := opslevel.TagCreateInput{
+				Key:   tagKey,
+				Value: tagValue,
+			}
+			if common.IsID(teamKey) {
+				input.Id = teamKey
+			} else {
+				input.Alias = teamKey
```
rocktavious, line 130: same as above

src/cmd/team.go
rauhut: It does work if we use the alias since on line [120](https://github.com/OpsLevel/cli/pull/72/files#diff-8becba81854ddbaceaee2adba092fff02e24b2d565d93a056623285ee25b7086R120) we're setting the type to be `TaggableResourceTeam`

Screen Shot 2022-11-11 at 1 11 17 PM

rocktavious: 😵‍💫  sorry - you are correct.

rocktavious (commented):
LGTM

eochoa260 (approved):
LGTM too

PR #78, opened 2022-12-14
Eo/add create repo grep

eochoa260:
Was able to create, get, and export a repo grep check.

### DIFF
--- a/.changes/unreleased/Dependency-20221216-071020.yaml
+++ b/.changes/unreleased/Dependency-20221216-071020.yaml
@@ -1,0 +1,3 @@
+kind: Dependency
+body: Bump github.com/opslevel/opslevel-go to v2022.12.16
+time: 2022-12-16T07:10:20.359077-06:00
--- a/.changes/unreleased/Feature-20221214-140633.yaml
+++ b/.changes/unreleased/Feature-20221214-140633.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: Add ability to create and export Repo Grep checks
+time: 2022-12-14T14:06:33.615436-06:00
--- a/src/cmd/check.go
+++ b/src/cmd/check.go
@@ -25,7 +25,7 @@ var getCheckCmd = &cobra.Command{
 		if isYamlOutput() {
 			common.YamlPrint(marshalCheck(*check))
 		} else {
-			common.PrettyPrint(check)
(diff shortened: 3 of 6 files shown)


rocktavious (approved):
Did you test export works on Circular's Repo Grep?  If so then LGTM!

rocktavious:
There appears to be a bug - when i test this out on circular with a repo grep check it doesn't print out the fields in JSON output but it does work in YAML output.  🤔

eochoa260:
> There appears to be a bug - when i test out `get check XXX` on circular with a repo grep check it doesn't print out the fields in JSON output but it does work in YAML output. 🤔

Just now seeing this but I'll look into it and see if I can figure out what's going on.  I only tested YAML output before.

eochoa260:
> > There appears to be a bug - when i test out `get check XXX` on circular with a repo grep check it doesn't print out the fields in JSON output but it does work in YAML output. 🤔
> 
> Just now seeing this but I'll look into it and see if I can figure out what's going on. I only tested YAML output before.

Looks like it was broken for other check types too.  Seems like running it through marshalCheck fixed it:
```
$ go run main.go get check Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpSZXBvR3JlcC8xODQy
{
  "Version": "1",
  "Kind": "repo_grep",
  "Spec": {
    "category": "performance",
    "directorySearch": false,
    "enabled": true,
    "fileContentsPredicate": {
      "Type": "exists",
      "Value": ""
    },
    "filePaths": [
      "**/go.mod"
    ],
    "filter": "test-filter",
    "level": "bronze",
    "name": "Test Repo Grep",
    "notes": "",
    "owner": "platform"
  }
}
$ go run main.go get check Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpSZXBvRmlsZS8xODQz        
{
  "Version": "1",
  "Kind": "repo_file",
  "Spec": {
    "category": "security",
    "directorySearch": false,
    "enabled": true,
    "fileContentsPredicate": {
      "Type": "equals",
      "Value": "postgres"
    },
    "filePaths": [
      "/src",
      "/test"
    ],
    "filter": "test-filter",
    "level": "silver",
    "name": "Hello World",
    "notes": "",
    "owner": "platform"
  }
}
```

rocktavious (approved):
Thanks for fixing a non-related bug

PR #84, opened 2023-01-20
Improve upon gaps with `teams` commands

rocktavious:


src/cmd/team.go
```diff
@@ -53,9 +64,9 @@ var createMemberCmd = &cobra.Command{
 
 var contactType string
 var createContactCmd = &cobra.Command{
-	Use:        "contact TEAM_ID|TEAM_ALIAS ADDRESS DISPLAYNAME",
+	Use:        "contact {TEAM_ID|TEAM_ALIAS} ADDRESS DISPLAYNAME",
 	Short:      "Add a contact to a team",
-	Long:       `Add a contact to a team`,
+	Example:    `opslevel create contact --type=slack my-team #general General`,
```
rocktavious, line 69: should we add more examples here?

eochoa260: Maybe one more wouldn't hurt.  Or having all the types in the help command but that might be harder to do.

eochoa260 (approved):
LGTM

src/cmd/team.go
rocktavious: Hmm ya i think i can do this

PR #89, opened 2023-02-01
Designing a new kind of cli command

rocktavious:
This allows for us to use the CLI like curl for any arbitrary graphql query.  While its not terribly simple it is more simple then curl AND it handles pagination and aggregation automatically.

There are a few samples in the command long description but to put one here

```bash
opslevel graphql -f owner="platform" -f tier="tier_1" --paginate -a=".account.services.nodes[]" -q='
query ($endCursor: String, $owner: String!, $tier: String!) {
  account {
    services(first: 1, after: $endCursor, ownerAlias: $owner, tierAlias: $tier) {
      nodes {
        name
        aliases
      }
      pageInfo {
        hasNextPage
        endCursor
      }
    }
  }
}'
```


src/cmd/graphql.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		log.Error().Err(err).Msg(msg)
+		os.Exit(1)
+	}
+}
+
+func convert(v string) (interface{}, error) {
+	if v == "-" {
+		reader := bufio.NewReader(os.Stdin)
+		return reader.ReadString('\n')
```
rocktavious, line 170: recurse here on convert so that if the stdin or a file are read in they still convert to a int or bool

src/cmd/graphql.go
```diff
@@ -0,0 +1,213 @@
...
+      }
+      pageInfo {
+        hasNextPage
+        endCursor
+      }
+    }
+  }
+}'
+
+opslevel graphql -f owner='platform' -f tier="tier_1" --paginate -a=".account.services.nodes[]" -q='
```
rocktavious, line 67: fix quotes

src/cmd/graphql.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/opslevel/opslevel-go/v2023"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var keyValueExp = regexp.MustCompile(`([\w-]+)=(.*)`)
```
rocktavious, line 17:
```suggestion
var keyValueExp = regexp.MustCompile(`([\w-]+)="(.*)"`)
```

src/cmd/graphql.go
```diff
@@ -0,0 +1,213 @@
...
+      }
+      pageInfo {
+        hasNextPage
+        endCursor
+      }
+    }
+  }
+}'
+
+opslevel graphql -f owner='platform' -f tier="tier_1" --paginate -a=".account.services.nodes[]" -q='
```
rocktavious, line 67:
```suggestion
opslevel graphql -f owner="platform" -f tier="tier_1" --paginate -a=".account.services.nodes[]" -q='
```

matthewbrahms (approved):
i think i understand at a high level what we are doing here, but it will take working with it a bit more to really get it all.  good work!

eochoa260 (approved):
LGTM

PR #95, opened 2023-03-03
fix graphql command and add another example

rocktavious:
after doing some more testing i found a bug with key=value parsing and this change makes it make the most sense.

I also added a really cool example of showing how to use the internal API to download the opslevel.yml file for a service

### DIFF
--- a/src/cmd/graphql.go
+++ b/src/cmd/graphql.go
@@ -14,7 +14,7 @@ import (
     "strings"
 )
 
-var keyValueExp = regexp.MustCompile(`([\w-]+)="(.*)"`)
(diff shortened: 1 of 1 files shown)


src/cmd/graphql.go
```diff
@@ -79,6 +79,15 @@ query ($endCursor: String, $owner: String!, $tier: String!) {
     }
   }
 }'
+
+opslevel graphql -f "id=XXXXXX" -H "GraphQL-Visibility=internal" -a=".account.configFile.yaml" -q='
```
eochoa260, line 83: 🔥

eochoa260 (approved):
LGTM

PR #97, opened 2023-03-12
Example of how to test the CLI

rocktavious:


taimoorgit:
Here is another snippet I have saved on my PC regarding testing. I came up with this when working on some preliminary codegen work a while ago:

```
// THIS is what we need to generate.
func CreateScorecardCommand(client *opslevel.Client) *cobra.Command {
    run := func(cmd *cobra.Command, args []string) {
        if client == nil {
            client = getClientGQL()
        }

        input, err := readResourceInput[opslevel.ScorecardInput]()
        cobra.CheckErr(err)
        res, err := client.CreateScorecard(*input)
        cobra.CheckErr(err)

        output, err := json.MarshalIndent(res, "", "    ")
        cobra.CheckErr(err)
        fmt.Println(string(output))
    }
    return NewCreateCmd[opslevel.ScorecardInput]("scorecard", nil, run)
}

// below is not generated, already exists in create.go or somewhere else.
func NewCreateCmd[T any](resName string, cmdAliases []string, run func(cmd *cobra.Command, args []string)) *cobra.Command {
    lowerName := strings.ToLower(resName)
    if run == nil {
        err := fmt.Errorf("NewCreateCmd: run arg for '%s' is nil", lowerName)
        panic(err)
    }

    return &cobra.Command{
        Aliases: cmdAliases,
        Use:     lowerName,
        Short:   fmt.Sprintf("create %s %s", grammar.IndefArticle(lowerName), lowerName),
        Long:    fmt.Sprintf("create %s %s", grammar.IndefArticle(lowerName), lowerName),
        Example: fmt.Sprintf(CreateStringExample, lowerName, getExample[T]()),
        Run:     run,
    }
}
```

It allows passing a client into any command that can be used to do things like evaluating what is contained in the request.

PR #103, opened 2023-03-23
Add domain commands

eochoa260:
Dependent on https://github.com/OpsLevel/opslevel-go/pull/178 getting merged first.

### DIFF
--- a/src/cmd/domain.go
+++ b/src/cmd/domain.go
@@ -1,11 +1,16 @@
 package cmd
 
 import (
+	"encoding/csv"
+	"encoding/json"
 	"fmt"
 	"github.com/creasty/defaults"
+	"github.com/opslevel/cli/common"
 	"github.com/opslevel/opslevel-go/v2023"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"os"
+	"strings"
 )
 
 var createDomainCmd = &cobra.Command{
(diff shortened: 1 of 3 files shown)


src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+			w := common.NewTabWriter("NAME", "ALIAS", "ID")
+			for _, item := range list {
+				fmt.Fprintf(w, "%s\t%s\t%s\t\n", item.Name, item.Id, strings.Join(item.Aliases, ","))
+			}
+			w.Flush()
+		}
+	},
+}
+
+var updateDomainCmd = &cobra.Command{
```
eochoa260, line 180: Getting a weird error on update, will share the details with team maturity tomorrow since I'm not sure what it's complaining about:  
```

Error: OpsLevel API Errors:
        - 'internal' {:entity_schema=>"must be present"}

exit status 1
```

eochoa260: Verified that I get the same error when using the graphiql explorer too.

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+	"github.com/creasty/defaults"
+	"github.com/opslevel/cli/common"
+	"github.com/opslevel/opslevel-go/v2023"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"os"
+	"strings"
+)
+
+var assignSystemCmd = &cobra.Command{
```
rocktavious, line 16: This command won't work when `opslevel create system` is registered.  We should chat about the design of this one.  I would expect the parent domain to be a parameter on `create system` or `update system` and this command is unncessary

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+
+var createDomainCmd = &cobra.Command{
+	Use:   "domain",
+	Short: "Create a domain",
+	Long: `Create a domain
+
+cat << EOF | opslevel create domain -f -
+name: "My Domain"
+description: "Hello World Domain"
+ownerId: "Z2lkOi8vb3BzbGV2ZWwvVGVhbS83NjY"
```
rocktavious, line 44: 🤮 - but I understand why.  I wonder if there is a better way for us to cover this up with the CLI?  Maybe with a special flag on the command `opslevel create domain --ownerType=group` or something?  Then the field can be an alias and we can lookup the group and fill it with the id.

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+		addErr := domain.AssignSystem(getClientGQL(), system)
+		cobra.CheckErr(addErr)
+		fmt.Printf("add system '%s' to domain '%s'\n", system, domain.Name)
+	},
+}
+
+var createDomainCmd = &cobra.Command{
+	Use:   "domain",
+	Short: "Create a domain",
+	Long: `Create a domain
```
rocktavious, line 39: cobra has a new field `Example` that you can use for examples.  Checkout the user.go for an example

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		key := args[0]
+		result, err := getClientGQL().GetDomain(key)
+		cobra.CheckErr(err)
+		common.WasFound(result.Id == "", key)
+		common.PrettyPrint(result)
+	},
+}
+
+var getDomainSystemCmd = &cobra.Command{
```
rocktavious, line 86: Any reason to not merge this logic into `get` and `list` domain?  This way it can be part of the JSON output of those commands.

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+	Short:      "Get details about a domain",
+	Long:       `Get details about a domain`,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID", "ALIAS"},
+	Run: func(cmd *cobra.Command, args []string) {
+		key := args[0]
+		result, err := getClientGQL().GetDomain(key)
+		cobra.CheckErr(err)
+		common.WasFound(result.Id == "", key)
+		common.PrettyPrint(result)
```
rocktavious, line 82: Does this take into account the different output types - json and yaml?  I'm pretty sure our other get commands have optional output formats.

src/cmd/domain.go
```diff
@@ -0,0 +1,233 @@
...
+			w := common.NewTabWriter("Name", "ID")
+			for _, item := range systems {
+				fmt.Fprintf(w, "%s\t%s\t\n", item.Name, item.Id)
+			}
+			w.Flush()
+		}
+	},
+}
+
+var getDomainTagCmd = &cobra.Command{
```
rocktavious, line 113: I don't understand the purpose of this command - tags should come through on the `get` and `list` commands data?  If they do not then we need to merge this logic.

rocktavious (changes requested):
Let me know if you want to talk through any feedback.

src/cmd/domain.go
eochoa260: Ah ok, yeah I was unsure about this one since I thought it might conflict but thought maybe it would use the right command based on arguments passed in.  But yeah, the create system and update system commands should take a parent input and do the same thing.  Is there a reason we added the AssignSystem mutation then?

src/cmd/domain.go
eochoa260: I didn't like this either but wasn't sure if there was an easy way to convert this?  Could we write something on the opslevel-go side that gets the team with the alias and then passes the ID to the input?

src/cmd/domain.go
eochoa260: I can add the optional output format for json and yaml.  Not sure it takes it into account here.

src/cmd/domain.go
eochoa260: I thought we were following the same pattern as groups and having separate commands to get the "child" resources.  But I can add the `ChildSystems` call in the get and list and add the systems to the domain object but we don't have a field for that so I thought we were going to leave it separate.  If we include it in the domain output would it not be better to just have a hydrate function like on services?  I thought we decided against this last time we talked about it though.

src/cmd/domain.go
eochoa260: Same reasoning as above.

rocktavious (approved):

PR #104, opened 2023-03-29
Mb/cli system create

matthewbrahms:


src/system
```diff
@@ -0,0 +1,6 @@
+name: "Test System"
```
rocktavious, line 1: I think this file wasn't meant to be committed

rocktavious (approved):
There just 1 local testing file that was commited that shouldn't be otherwise its good

PR #129, opened 2023-06-19
PrettyPrint no longer escapes html

alan-opslevel:
- create `json.Encoder` to customize `SetEscapeHTML` and `SetIndent` to print `rawNotes` without escaping these following chars:
  - "<", ">", "&", U+2028, and U+2029 are escaped to "\u003c","\u003e", "\u0026", "\u2028", and "\u2029"

this misbehaviour was discovered during investigation from this [ticket](https://gitlab.com/jklabsinc/OpsLevel/-/issues/7693)

### DIFF
--- a/.changes/unreleased/Bugfix-20230619-172139.yaml
+++ b/.changes/unreleased/Bugfix-20230619-172139.yaml
@@ -1,0 +1,3 @@
+kind: Bugfix
+body: PrettyPrint no longer escapes html
+time: 2023-06-19T17:21:39.366408-04:00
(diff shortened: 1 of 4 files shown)


matthewbrahms:
I think the only thing Kyle would ask would be if we had time to write a test that would cover this in happening in the futur?. If there isn't time, we can circle back around to that at a later time. 

cheers.

alan-opslevel:
> @alan-opslevel I think the only thing Kyle would ask would be if we had time to write a test that would cover this in happening in the futur?. If there isn't time, we can circle back around to that at a later time.

will add a test

matthewbrahms (approved):
thanks @alan-opslevel

PR #131, opened 2023-07-06
add Taskfile, gofmt code, fix lint issues

davidbloss:
Added Taskfile.yml that formats, lints, and tests code.
Added a boat load of error checks and some other linter fixes.

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -20,7 +20,9 @@ func init() {
 	rootCmd.AddCommand(createCmd)
 
 	createCmd.PersistentFlags().StringVarP(&createDataFile, "file", "f", "-", "File to read data from. If '.' then reads from './data.yaml'. Defaults to reading from stdin.")
-	viper.BindPFlags(createCmd.Flags())
+	if err := viper.BindPFlags(createCmd.Flags()); err != nil {
+		cobra.CheckErr(err)
+	}
 }
 
 func hasStdin() bool {
@@ -58,7 +60,9 @@ func readCreateConfigFile() {
 		if createDataFile == "-" {
 			viper.SetConfigType("yaml")
 			if hasStdin() {
-				viper.ReadConfig(os.Stdin)
+				if err := viper.ReadConfig(os.Stdin); err != nil {
+					cobra.CheckErr(err)
+				}
 			}
 			return
 		} else if createDataFile == "." {
@@ -75,5 +79,7 @@ func readCreateConfigFile() {
 		viper.AddConfigPath(".")
 		viper.AddConfigPath(home)
 	}
-	viper.ReadInConfig()
+	if err := viper.ReadInConfig(); err != nil {
+		cobra.CheckErr(err)
+	}
 }
--- a/src/cmd/deploy.go
+++ b/src/cmd/deploy.go
@@ -2,9 +2,10 @@ package cmd
 
 import (
 	"encoding/json"
-	"github.com/opslevel/opslevel-go/v2023"
 	"os"
 	"time"
+
+	"github.com/opslevel/opslevel-go/v2023"
 
 	"github.com/creasty/defaults"
 	git "github.com/go-git/go-git/v5"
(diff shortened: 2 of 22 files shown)


src/cmd/create.go
```diff
@@ -20,7 +20,9 @@ func init() {
 	rootCmd.AddCommand(createCmd)
 
 	createCmd.PersistentFlags().StringVarP(&createDataFile, "file", "f", "-", "File to read data from. If '.' then reads from './data.yaml'. Defaults to reading from stdin.")
-	viper.BindPFlags(createCmd.Flags())
+	if err := viper.BindPFlags(createCmd.Flags()); err != nil {
```
rocktavious, line 23: Not sure which is more idomatic go but it could also be like this

```suggestion
	cobra.CheckErr(viper.BindPFlags(createCmd.Flags()))
```

since `cobra.CheckErr` already does that if statement checking for nil.

src/cmd/create.go
```diff
@@ -75,5 +79,7 @@ func readCreateConfigFile() {
 		viper.AddConfigPath(".")
 		viper.AddConfigPath(home)
 	}
-	viper.ReadInConfig()
+	if err := viper.ReadInConfig(); err != nil {
```
rocktavious, line 82: same as above possibly

src/cmd/create.go
```diff
@@ -58,7 +60,9 @@ func readCreateConfigFile() {
 		if createDataFile == "-" {
 			viper.SetConfigType("yaml")
 			if hasStdin() {
-				viper.ReadConfig(os.Stdin)
+				if err := viper.ReadConfig(os.Stdin); err != nil {
```
rocktavious, line 63: same as above possibly

src/cmd/deploy.go
```diff
@@ -101,23 +102,57 @@ func init() {
...
-	viper.BindEnv("dry-run", "OPSLEVEL_DRY_RUN", "OL_DRY_RUN")
-	viper.BindEnv("git-path", "OPSLEVEL_GIT_PATH", "OL_GIT_PATH")
-	viper.BindEnv("deploy-number", "OPSLEVEL_DEPLOY_NUMBER", "OL_DEPLOY_NUMBER")
-	viper.BindEnv("deploy-url", "OPSLEVEL_DEPLOY_URL", "OL_DEPLOY_URL")
-	viper.BindEnv("dedup-id", "OPSLEVEL_DEDUP_ID", "OL_DEDUP_ID")
-	viper.BindEnv("deployer-name", "OPSLEVEL_DEPLOYER_NAME", "OL_DEPLOYER_NAME")
-	viper.BindEnv("deployer-email", "OPSLEVEL_DEPLOYER_EMAIL", "OL_DEPLOYER_EMAIL")
-	viper.BindEnv("commit-sha", "OPSLEVEL_COMMIT_SHA", "OL_COMMIT_SHA")
-	viper.BindEnv("commit-message", "OPSLEVEL_COMMIT_MESSAGE", "OL_COMMIT_MESSAGE")
+	if err := viper.BindPFlags(deployCreateCmd.Flags()); err != nil {
```
rocktavious, line 105: hmm - do we know when this binding of flags would fail?  I'm concerned this add a TON of boilerplate for no real value.  If this is worth it then it would make make sense to make a wrapper method to handle the error checking

```
func BindEnv(key string, envs string...) {
	cobra.CheckErr(viper.BindEnv(key, ...envs))
}
```

src/cmd/document.go
```diff
@@ -29,7 +30,7 @@ opslevel create document my-service -r services -t openapi -i xxxxx -f swagger.j
 		integrationID, err := flags.GetString("integration-id")
 		cobra.CheckErr(err)
 		integrationURL := fmt.Sprintf("integrations/document/%s/%s/%s/%s", integrationID, resourceType, serviceAlias, documentType)
-		fileContents, err := ioutil.ReadFile(createDataFile)
+		fileContents, err := os.ReadFile(createDataFile)
```
rocktavious, line 33: TIL

src/cmd/domain.go
```diff
@@ -90,9 +91,14 @@ opslevel list domain
 			common.JsonPrint(json.MarshalIndent(list, "", "    "))
 		} else if isCsvOutput() {
 			w := csv.NewWriter(os.Stdout)
-			w.Write([]string{"NAME", "ID", "ALIASES"})
+			if err := w.Write([]string{"NAME", "ID", "ALIASES"}); err != nil {
+				panic(err)
```
rocktavious, line 95: its recommended to always `cobra.CheckErr` when inside a cobra command because it properly sets the exit code

src/cmd/domain.go
```diff
@@ -90,9 +91,14 @@ opslevel list domain
...
+			}
 			for _, item := range list {
-				w.Write([]string{item.Name, string(item.Id), strings.Join(item.Aliases, "/")})
+				err := w.Write([]string{item.Name, string(item.Id), strings.Join(item.Aliases, "/")})
+				if err != nil {
```
rocktavious, line 99: I'm not sure we actually care about the error here.  Why not just continue letting it write more lines?   Unfortunately logging errors here will screw up the output most likely too.  I think this is an OK error to skip over.

src/cmd/graphql.go
```diff
@@ -1,29 +1,32 @@
...
+	"github.com/spf13/cobra"
 )
 
-var keyValueExp = regexp.MustCompile(`([\w-]+)=(.*)`)
-var usesPaginationExp = regexp.MustCompile(`\$endCursor:\WString`)
-var hasNextPageExp = regexp.MustCompile(`"hasNextPage":([\w]+)`)
-var endCursorExp = regexp.MustCompile(`"endCursor":\"([\w]+)\"`)
+var (
+	keyValueExp = regexp.MustCompile(`([\w-]+)=(.*)`)
+	// var usesPaginationExp = regexp.MustCompile(`\$endCursor:\WString`)
```
rocktavious, line 20: feels like a bug that this is no used.  🤔

src/cmd/graphql.go
```diff
@@ -89,138 +92,138 @@ query ($id: ID!){
...
+			} else {
+				hasNextPage = false
+			}
+		}
+
+		json, err := json.Marshal(output)
+		handleErr("error marshaling output to json", err)
+
+		fmt.Println(string(json))
+	},
```
rocktavious, line 158: Where there any changes in here besides spaces?  Its unclear to me.

src/cmd/group.go
```diff
@@ -278,13 +279,8 @@ var deleteGroupCmd = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
 		key := args[0]
-		if common.IsID(key) {
-			err := getClientGQL().DeleteGroup(key)
-			cobra.CheckErr(err)
-		} else {
-			err := getClientGQL().DeleteGroupWithAlias(key)
-			cobra.CheckErr(err)
-		}
+		err := getClientGQL().DeleteGroup(key)
```
rocktavious, line 282: Does this stay functionality the same?

src/cmd/policy.go
```diff
@@ -184,15 +184,17 @@ Examples:
...
 		cobra.CheckErr(err)
 
 		if outputFilePath == "-" {
 			fmt.Println(string(b))
 		} else {
 			main := newFile(outputFilePath, false)
 			defer main.Close()
-			main.WriteString(string(b))
+			if _, err := main.WriteString(string(b)); err != nil {
+				log.Error().Err(err).Msg("")
```
rocktavious, line 196: We need to ensure this will not mess up the final output that the Ruby app might be expecting.  This command is used by opslevel-runner service detection jobs that the monolith uses.

Also we should always add a message to the `.Msg` at least describing the expected errors this might throw so in the future if we see that error message we can do a code search to find this error line.

src/cmd/team.go
```diff
@@ -344,7 +347,9 @@ var deleteTeamTagCmd = &cobra.Command{
 		} else {
 			for _, tag := range result.Tags.Nodes {
 				if tagKey == tag.Key {
-					getClientGQL().DeleteTag(tag.Id)
+					if err := getClientGQL().DeleteTag(tag.Id); err != nil {
+						cobra.CheckErr(err)
+					}
```
rocktavious, line 352:
```suggestion
					getClientGQL().DeleteTag(tag.Id)
					cobra.CheckErr(err)
```

src/cmd/terraform.go
```diff
@@ -27,9 +28,9 @@ func init() {
 }
 
 func newFile(filename string, makeExecutable bool) *os.File {
-	path := fmt.Sprintf("%s", filename)
+	path := filename
```
rocktavious, line 31: this seems silly now - not sure why it was like this but we can probably refactor away the `path` variable and just use `filename` everywhere in this function.

src/cmd/terraform.go
```diff
@@ -93,7 +94,13 @@ func runExportTerraform(cmd *cobra.Command, args []string) {
 provider "opslevel" {
 }
 `)
-	bash.WriteString("#!/bin/sh\n\n")
+	if err != nil {
+		panic(err)
```
rocktavious, line 98: this is inside a cobra command so we should use `cobra.CheckErr`

src/cmd/create.go
davidbloss: Great point! This is much simpler and achieves the same outcome 👍

src/cmd/deploy.go
davidbloss: Agreed with cutting down the redundancy. I plan on looking at this again in a follow up PR and likely writing a wrapper function like your example

src/cmd/domain.go
davidbloss: Most definitely 👍

src/cmd/domain.go
davidbloss: ☝️ keeping this in mind for a follow up PR. Really I'm inclined to fully build the string then do the actual write operation once. Also

src/cmd/graphql.go
davidbloss: I think there was one place that changed a for loop to an append(). But this is a mess. These changes will be introduced in a follow up PR

src/cmd/group.go
davidbloss: Yes, [DeleteGroupWithAlias](https://pkg.go.dev/github.com/opslevel/opslevel-go/v2023@v2023.5.18#Client.DeleteGroup) is basically a wrapper that uses `DeleteGroup` under the hood

src/cmd/policy.go
davidbloss: Good catch! Looks like I was rushed this part a bit.

davidbloss:
This PR is too busy. I'm breaking the updates suggested here into multiple smaller PRs to follow. Thanks for the comments @rocktavious 🎉  I will be referring back to them for the next PRs

PR #134, opened 2023-07-14
Add alias create and delete commands

rocktavious:
Infrastructure resources have no way to set the alias on them but it seems that `aliasCreate` can do it so i'm adding these commands to the CLI so you can manage aliases on infrastructure resources as well as other things.


davidbloss (approved):
Looks good to me 👍

PR #139, opened 2023-07-25
Create and update filters using YAML

taimoor-at-opslevel:
## Problem

We can't create filters with predicates using YAML and can't update filters at all.

## Solution

Update `create filter -f ...` to allow a file argument, add `update filter $FILTER_ID -f ...` command.

## Tophatting

Setup: I added these 2 lines before the graphql calls and created 2 yaml files for create/update. I added these 2 lines before the graphql call to ensure files are being read correctly.

Can create a filter using YAML file:

```
$ go run main.go create filter -f create_filter.yaml
&{Name:Tier 1 apps using RDS Predicates:[{Key:tier_index KeyData: Type:equals Value:1} {Key:tags KeyData:db Type:equals Value:rds}] Connective:and}
```

Can update a filter using YAML file, by providing `id` as the first argument to the command:

```
$ go run main.go update filter Z2lkOi8vb3BzbGV2ZWwvRmlsdGVyLzIzNTk -f update_filter.yaml
&{Id:Z2lkOi8vb3BzbGV2ZWwvRmlsdGVyLzIzNTk Name:Tier 2 apps using RDS Predicates:[{Key:tier_index KeyData: Type:equals Value:2} {Key:tags KeyData:db Type:equals Value:dynamo}] Connective:and}
```

Tested out the example in the CLI live, real create:

```
$ cat << EOF | go run main.go create filter -f -
name: "Tier 1 apps using RDS"
connective: "and"
predicates:
  - key: "tier_index"
    type: "equals"
    value: "1"
  - key: "tags"
    keyData: "db"
    type: "equals"
    value: "rds"
EOF
Z2lkOi8vb3BzbGV2ZWwvRmlsdGVyLzIzNzI
```

Real update:

```
$ cat << EOF | go run main.go update filter Z2lkOi8vb3BzbGV2ZWwvRmlsdGVyLzIzNzI -f -
name: "Updated filter"
connective: "or"
predicates:
  - key: "tier_index"
    type: "equals"
    value: "4"
  - key: "tags"
    keyData: "db"
    type: "equals"
    value: "sqlite"
EOF
{
    "Connective": "or",
    "HtmlURL": "https://app.opslevel.com/filters/2372",
    "Id": "Z2lkOi8vb3BzbGV2ZWwvRmlsdGVyLzIzNzI",
    "Name": "Updated filter",
    "Predicates": [
        {
            "key": "tier_index",
            "type": "equals",
            "value": "4"
        },
        {
            "key": "tags",
            "keyData": "db",
[...]

PR #141, opened 2023-07-26
Bugfix for reading YAML files on update

taimoor-at-opslevel:
Blocks https://github.com/OpsLevel/cli/pull/139

## Problem

Bug: `opslevel update $RESOURCE -f $CONFIG.yaml` will not read the input YAML file, unless you pass `-`. This happens for resources like `domains` which don't have a separate input file reading function call. [See example in domain.go](https://github.com/OpsLevel/cli/blob/842ae0260021d1bfc45ba5e63282749a29806c77/src/cmd/domain.go#L141) where creating a domain using a YAML file works but updating one does not.

`-f -` (stdin) is able to work because of persistent flags set in [create.go](https://github.com/OpsLevel/cli/blob/842ae0260021d1bfc45ba5e63282749a29806c77/src/cmd/create.go#L22) and [update.go.](https://github.com/OpsLevel/cli/blob/842ae0260021d1bfc45ba5e63282749a29806c77/src/cmd/update.go#L21)

## Solution

Merge file reading functions from create.go and update.go into input.go, share `dataFile` variable between them.

## Tophatting

Setup: I added these 2 lines before the graphql calls in domain.go and created a file called create_domain.yaml and update_domain.yaml.

```
fmt.Printf("%s\n%s\n%s\n%s\n", *input.Description, *input.Name, *input.Note, *input.Owner)
os.Exit(1)
```

Can read yaml input for create:

```
$ go run main.go create domain -f create_domain.yaml
Hello World Domain
My Domain
Additional details
Z2lkOi8vb3BzbGV2ZWwvVGVhbS83NjY
```

Can read yaml input for update:

```
$ go run main.go update domain reddit.com -f update_domain.yaml
Updated Hello World Domain
My Updated Domain
This was updated
Z2lkOi8vb3BzbGV2ZWwvVGVhbS83NjY
```

### DIFF
--- a/src/cmd/create.go
+++ b/src/cmd/create.go
@@ -1,14 +1,9 @@
 package cmd
 
 import (
-	"io"
-	"os"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
-
-var createDataFile string
 
 var createCmd = &cobra.Command{
 	Use:   "create",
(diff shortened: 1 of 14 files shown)


src/cmd/input.go
```diff
@@ -0,0 +1,66 @@
...
+
+		data, err := io.ReadAll(file)
+		if err != nil {
+			return nil, err
+		}
+		return data, nil
+	}
+}
+
+func readInputConfig() {
```
taimoor-at-opslevel, line 43: I named this function `InputFile` rather than `InputConfigFile` so it's easier to differentiate with `readInputFile`.

This handles viper config, the latter handles an actual file (or stdin)

src/cmd/create.go
```diff
@@ -1,15 +1,10 @@
 package cmd
 
 import (
-	"io"
```
taimoor-at-opslevel, line 4: Default Go formatter re-ordered inputs when I was updating references, we can remove these if not wanted.

src/cmd/update.go
```diff
@@ -1,46 +1,19 @@
...
-	Long:  "Update resources in OpsLevel",
+	Short: "Update resources or events from a file or stdin",
+	Long:  "Update resources or events from a file or stdin",
 }
 
 func init() {
 	rootCmd.AddCommand(updateCmd)
 
-	updateCmd.PersistentFlags().StringVarP(&updateDataFile, "file", "f", "-", "File to read update from. If '.' then reads from './data.yaml'. Defaults to reading from stdin.")
-	viper.BindPFlags(createCmd.Flags())
```
taimoor-at-opslevel, line 22: I think the reference to `createCmd` here is a mistake - we want this flag binded globally if we're using update.

rocktavious (approved):

taimoor-at-opslevel:
https://github.com/OpsLevel/cli/pull/147 is kind of a follow up to this.

PR #148, opened 2023-08-03
Add support for updating repo grep checks

taimoor-at-opslevel:
**Relies on https://github.com/OpsLevel/opslevel-go/pull/229**

## Problem

Cannot update repo grep checks (and checks in general). 

No examples for creating a check.

We are have unreliable/unneeded functions in input.go.

## Solution

1. Add logic for updating a check
1. Separate the logic for creating and updating a check into 2 files. The files are the same, the only differences are just `sed s/CreateInput/UpdateInput/g` TODO: change this, probably can combine both files into one. 
1. Add a `--noninteractive flag` flag to prevent prompt UI from showing up. (See code comment)
1. Remove unreliable/unneeded functiosn in input.go that were only being used by dependency.go.

## Tophatting

**Can create, update, get a repo grep check**

```
$ cat << EOF | go run main.go create check -f -
Version: "1"
kind: "repo_grep"
spec:
  name: "new repo grep check"
  enabled: false
  category: "misc"
  level: "silver"
  notes: "some notes"
  directorySearch: false
  filePaths:
  - "Taskfile.yml"
  fileContentsPredicate:
    type: "exists"
    value: ""
EOF
3:50PM WRN running noninteractively since using heredoc
3:50PM WRN no value supplied for field 'owner'
3:50PM WRN no value supplied for field 'filter'
Created Check 'new repo grep check' with id 'Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpSZXBvR3JlcC8xMjY2Ng'
```

```
$ cat << EOF | go run main.go update check $CHECK_ID -f -
Version: "1"
kind: "repo_grep"
spec:
  name: "updated repo grep check"
  enabled: true
  fileContentsPredicate:
    type: "does_not_match_regex"
    value: "gofmt"
EOF
3:52PM WRN running noninteractively since using heredoc
3:52PM WRN no value supplied for field 'owner'
3:52PM WRN no value supplied for field 'filter'
# output truncated
```

```
$ go run main.go get check $CHECK_ID
{
  "Version": "1",
  "Kind": "repo_grep",
  "Spec": {
    "category": "misc",
    "directorySearch": false,
    "enabled": true,
    "fileContentsPredicate": {
      "Type": "does_not_match_regex",
      "Value": "gofmt"
    },
    "filePaths": [
      "Taskfile.yml"
    ],
    "level": "silver",
    "name": "updated repo grep check",
    "notes": "some notes"
  }
}

```

**Can still create a dependency after input.go change**

```
# a printf and exit call was added to demonstrate that ripping out the functions from input.go works.
$cat << EOF | go run main.go create service dependency -f -
source: my-service-alias
target: my-target-alias
notes: |
  A new service depdency
EOF
&{Key:{Service:{Id: Alias:my-service-alias} DependsOn:{Id: Alias:my-target-alias}} Notes:A new service depdency
}%                                                                                                                                                                                                                    
```

**Can create, update, get a generic/custom event type check**

```
$ cat check_create_generic.yaml
Version: "1"
kind: "generic"
spec:
  name: "new custom event check"
  enabled: false
  category: "misc"
  level: "silver"
  notes: "some notes"
  serviceSelector: ".alias"
  successCondition: ".open | all(.age < 14)"%                                                                                                                                                                         
$ cat check_update_generic.yaml
Version: "1"
kind: "generic"
spec:
  name: "updated custom event check"
  enabled: false
  category: "misc"
  level: "bronze"
  notes: "this was updated"
  serviceSelector: ".alias"
  successCondition: ".open | all(.age < 365)"%                                                                                                                                                                        
```

```
$ go run main.go create check -f check_create_generic.yaml
✔ Platform
✔ None
✔ End Of Life
Created Check 'new custom event check' with id 'Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpHZW5lcmljLzEyNzQ0'
```

```
$ go run main.go update check Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpHZW5lcmljLzEyNzQ0 -f check_update_generic.yaml
# output truncated
```

```
$ go run main.go get check Z2lkOi8vb3BzbGV2ZWwvQ2hlY2tzOjpHZW5lcmljLzEyNzQ0
{
  "Version": "1",
  "Kind": "generic",
  "Spec": {
    "category": "misc",
    "enabled": false,
    "filter": "uses-ruby-on-rails",
    "integration": "generic-end-of-life",
    "level": "bronze",
    "message": "",
    "name": "updated custom event check",
    "notes": "this was updated",
    "owner": "platform",
    "serviceSelector": ".alias",
    "successCondition": ".open | all(.age < 365)"
  }
}
```


src/cmd/check.go
```diff
@@ -54,22 +56,74 @@ var listCheckCmd = &cobra.Command{
...
+    type: "exists"
+    value: ""
+EOF`,
 	Run: func(cmd *cobra.Command, args []string) {
-		usePrompts := !hasStdin()
-		input, err := readCheckCreateInput()
+		input, err := readCheckInput()
 		cobra.CheckErr(err)
+		usePrompts := true
+		if cmd.Flag("noninteractive").Changed {
```
taimoor-at-opslevel, line 81: I added a `--noninteractive` flag for when users want to skip using the prompt UI, this is useful for automations. `hasStdin()` isn't a reliable function, see https://github.com/OpsLevel/cli/pull/147

src/cmd/check.go
```diff
@@ -54,22 +56,74 @@ var listCheckCmd = &cobra.Command{
...
+			usePrompts = false
+		} else if dataFile == "-" {
```
taimoor-at-opslevel, line 83: Using a heredoc without disabling prompt UI results in this error message, because the input is sent to the prompt. Here's the error message I was getting without this `else if`:

```
Error: ^D
exit status 1
```

src/cmd/check.go
```diff
@@ -54,22 +56,74 @@ var listCheckCmd = &cobra.Command{
...
+  fileContentsPredicate:
+    type: "does_not_match_regex"
+    value: "gofmt"
+EOF`,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
+	Run: func(cmd *cobra.Command, args []string) {
+		input, err := readCheckInput()
+		cobra.CheckErr(err)
+		input.Spec["id"] = *opslevel.NewID(args[0])
```
taimoor-at-opslevel, line 113: The actual mutation inputs used by opslevel-go are created by inserting all the key-value pairs from `Spec` into a graphql request. 

We can use the same input type for `create` and `update` by simply adding the ID here, similar to how we are working with filters. https://github.com/OpsLevel/cli/blob/7a230568bc3dd94545c8b3cb57be0b610c06f2b2/src/cmd/filter.go#L99-L107

src/cmd/check.go
```diff
@@ -114,12 +167,16 @@ func (self *CheckCreateType) resolveCategoryAliases(client *opslevel.Client, use
 		}
 		self.Spec["categoryId"] = category.Id
 	} else {
+		if self.IsUpdateInput() {
```
taimoor-at-opslevel, line 170: If the request is to update a check and prompt UI isn't being used, if they didn't specify a value for the aliases (like a category or level or integration) don't add that to the request. (This makes the mutation not change that field)

src/cmd/check.go
```diff
@@ -302,95 +267,25 @@ func (self *CheckCreateType) resolveIntegrationAliases(client *opslevel.Client,
...
-		output, err = clientGQL.CreateCheckAlertSourceUsage(*input.AsAlertSourceUsageCreateInput())
-
-	case opslevel.CheckTypeGitBranchProtection:
-		output, err = clientGQL.CreateCheckGitBranchProtection(*input.AsGitBranchProtectionCreateInput())
+// Serialization
 
-	case opslevel.CheckTypeServiceDependency:
-		output, err = clientGQL.CreateCheckServiceDependency(*input.AsServiceDependencyCreateInput())
-	}
+func toJson(data map[string]interface{}) []byte {
```
taimoor-at-opslevel, line 281: Diff looks confusing here. This function is unchanged, I just moved it down.

src/cmd/check.go
taimoor-at-opslevel: This change could break things if the user upgrades their CLI version without realizing this flag was added and the old check was removed for being unreliable.

rocktavious: Lets make sure we add a changie entry for this calling out this breaking change and what users must do to get the old behavior back.

src/cmd/check.go
rocktavious: ahhh neat!

src/cmd/check.go
rocktavious: Ya this is legacy from before Evan joined and showed us the way to make input types not unique between create and update.

src/cmd/check.go
```diff
@@ -85,19 +139,18 @@ var deleteCheckCmd = &cobra.Command{
 }
 
 func init() {
+	checkCreateCmd.Flags().Bool("noninteractive", false, "don't use promptui - all linked fields must be defined in yaml")
```
rocktavious, line 142:
```suggestion
	checkCreateCmd.Flags().Bool("noninteractive", false, "turns off automated prompts for fields missing data in the spec")
```

src/cmd/check.go
```diff
@@ -85,19 +139,18 @@ var deleteCheckCmd = &cobra.Command{
...
 	createCmd.AddCommand(checkCreateCmd)
+	checkUpdateCmd.Flags().Bool("noninteractive", false, "don't use promptui - all linked fields must be defined in yaml")
```
rocktavious, line 144: same documentation change as right above
```suggestion
	checkUpdateCmd.Flags().Bool("noninteractive", false, "turns off automated prompts for fields missing data in the spec")
```

.changes/unreleased/Bugfix-20230810-114939.yaml
```diff
@@ -0,0 +1,4 @@
+kind: Bugfix
+body: 'BREAKING CHANGE: to disable interactive UI pass the --noninteractive flag when
```
rocktavious, line 2: @taimoor-at-opslevel - thinking about this more and reading it like this - i think we maybe have this backwards.  It feels like the default should be "non-interactive".  Thoughts?

rocktavious (approved):
Just 1 changie file change the its good

rocktavious:
To make the tests pass you need to commit the update to the submodule AND uncomment the `replace` stanza at the bottom of the go.mod

PR #149, opened 2023-08-08
Db/add workspace task

davidbloss:
Add `workspace` task to Taskfile to make `cli` <--> `opslevel-go` development setup easier(?) and more apparent.
Add `testcov` task to Taskfile for simpler CI and for easier parity for local testing.

I added `go.work*` to `.gitignore` since `go.work` only works if the `opslevel-go` submodule has been pulled in. This is handled by `task workspace` in CI testing and locally - and can be safely ignored in all other situations.


rocktavious (approved):
LGTM

PR #152, opened 2023-08-11
Add support for Webhook Actions and Trigger Definitions

taimoor-at-opslevel:
## Summary

Allow CLI users to create webhook actions and trigger definitions!

### Actions

```
$ go run main.go create action -f action_create.yaml
created webhook action: Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6V2ViaG9va0FjdGlvbi8yMjY
```

```
$ go run main.go update action Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6V2ViaG9va0FjdGlvbi8yMjY -f action_update.yaml
{
    "Aliases": [],
    "Id": "Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6V2ViaG9va0FjdGlvbi8yMjY",
    "Description": "Pages the oncall and creates an incident",
    "LiquidTemplate": "",
    "Name": "Page The On Call",
    "Headers": "{\"accept\":\"application/vnd.pagerduty+json;version=2\",\"authorization\":\"Token token=XXXXXXXXXXXXX\",\"from\":\"someone-else@example.com\"}",
    "HTTPMethod": "POST",
    "WebhookURL": "https://api.pagerduty.com/incidents"
}                                                                                                                                                                                  
```

### Trigger definitions

```
$ go run main.go create trigger-definition -f trigger_create.yaml
created trigger definition: Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6VHJpZ2dlckRlZmluaXRpb24vMjAx
```

```
$ go run main.go update trigger-definition Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6VHJpZ2dlckRlZmluaXRpb24vMjAx -f trigger_update.yaml
{
    "Action": {
        "Aliases": [],
        "Id": "Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6V2ViaG9va0FjdGlvbi85Mw"
    },
    "Aliases": [],
    "Description": "Pages the On Call via PagerDuty",
    "Filter": {
        "Id": null,
        "Name": ""
    },
    "Id": "Z2lkOi8vb3BzbGV2ZWwvQ3VzdG9tQWN0aW9uczo6VHJpZ2dlckRlZmluaXRpb24vMjAx",
    "ManualInputsDefinition": "version: 1\ninputs:\n  - identifier: IncidentTitle\n    displayName: Title\n    description: Title of the incident to trigger\n    type: text_input\n    required: true\n    maxLength: 60\n    defaultValue: Service Incident Manual Trigger\n  - identifier: IncidentDescription\n    displayName: Incident Description\n    description: The description of the incident\n    type: text_area\n    required: true\n",
    "Name": "Page The On Call",
    "Owner": {
        "Alias": "platform",
        "Id": "Z2lkOi8vb3BzbGV2ZWwvVGVhbS8xNjc4"
    },
    "Published": false,
    "Timestamps": {
        "createdAt": "2023-08-11T19:06:20.750274Z",
        "updatedAt": "2023-08-11T19:08:03.356306Z"
    },
    "AccessControl": "service_owners",
    "ResponseTemplate": "{% if response.status \u003e= 200 and response.status \u003c 300 %}\n## Congratulations!\nYour request for {{ service.name }} has succeeded. See the incident here: {{response.body.incident.html_url}}\n{% else %}\n## Oops something went wrong!\nPlease contact [{{ action_owner.name }}]({{ action_owner.href }}) for more help.\n{% endif %}",
    "EntityType": "SERVICE"
}
```

### DIFF
--- a/src/cmd/action.go
+++ b/src/cmd/action.go
@@ -66,9 +66,9 @@ var listActionCmd = &cobra.Command{
 		if isJsonOutput() {
 			common.JsonPrint(json.MarshalIndent(list, "", "    "))
 		} else {
-			w := common.NewTabWriter("ID", "NAME")
+			w := common.NewTabWriter("ID", "NAME", "HTTP_METHOD", "WEBHOOK_URL")
 			for _, item := range list {
-				fmt.Fprintf(w, "%s\t%s\t\n", item.Id, item.Name)
+				fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", item.Id, item.Name, item.HTTPMethod, item.WebhookURL)
 			}
 			w.Flush()
 		}
--- a/src/cmd/action.go
+++ b/src/cmd/action.go
@@ -1,0 +1,145 @@
+package cmd
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/creasty/defaults"
+	"github.com/opslevel/opslevel-go/v2023"
+
+	"github.com/opslevel/cli/common"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+// TODO: examples for create, update
+
+var actionType string
+
+var createActionCmd = &cobra.Command{
+	Use:   "action --type=$ACTION_TYPE",
+	Short: "Create an action",
+	Long:  "Create an action",
+	Example: `
+cat << EOF | opslevel create action --type=webhook -f -
+...
+EOF`,
+	Run: func(cmd *cobra.Command, args []string) {
+		switch actionType {
+		case "webhook":
+			input, err := readWebhookActionInput()
+			cobra.CheckErr(err)
+			result, err := getClientGQL().CreateWebhookAction(*input)
+			cobra.CheckErr(err)
+			fmt.Printf("created webhook action: %s\n", result.Id)
+		default:
+			err := errors.New("unknown action type: " + args[0])
+			cobra.CheckErr(err)
+		}
+	},
+}
+
+var getActionCmd = &cobra.Command{
+	Use:        "action ID",
+	Short:      "Get details about an action",
+	Long:       "Get details about an action",
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
+	Run: func(cmd *cobra.Command, args []string) {
+		key := args[0]
+		action, err := getClientGQL().GetCustomAction(*opslevel.NewIdentifier(key))
+		cobra.CheckErr(err)
+		common.PrettyPrint(action)
+	},
+}
+
+var listActionCmd = &cobra.Command{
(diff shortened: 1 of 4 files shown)


src/cmd/action.go
```diff
@@ -0,0 +1,171 @@
...
+			fmt.Printf("created webhook action: %s\n", result.Id)
+		default:
+			err := errors.New("unknown action type: " + args[0])
+			cobra.CheckErr(err)
+		}
+	},
+}
+
+var getActionCmd = &cobra.Command{
+	Use:        "action ID",
```
rocktavious, line 64:
```suggestion
	Use:        "action ID|ALIAS",
```
I think its safe to use ID or Alias on this command

src/cmd/action.go
```diff
@@ -0,0 +1,171 @@
...
+			for _, item := range list {
+				fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", item.Id, item.Name, item.HTTPMethod, item.WebhookURL)
+			}
+			w.Flush()
+		}
+	},
+}
+
+var updateActionCmd = &cobra.Command{
+	Use:   "action ID",
```
rocktavious, line 99: If Get supports `ID or ALIAS` but Update does not - that feels weird.  Is this true?

Maybe we need to paper over this in the CLI - lookup the action by Alias?

Not sure.

rocktavious (approved):
Overall it think its good to ship but had some questions about the end user experience of using IDs vs Aliases.  It maybe is a miss in the API but maybe there is a way for us to paper over it is we can truely work with Aliases instead of IDs.  

If i'm missing some sort of context that makes this impossible then lets ship it.

src/cmd/action.go
taimoor-at-opslevel: I noticed something interested with the GraphQL API when looking into this.

The mutation `CustomActionsWebhookActionUpdateInput!` doesn't have an ALIAS as a parameter at all, only ID. Same with `CustomActionsWebhookActionCreateInput!`.

The query for `CustomActionsExternalAction!` does take an `IdentifierInput!` where you can pass an ALIAS or ID. But since you can never create or update an ALIAS into an action, there's no point in using it, since there would never be an Action object where the `Aliases` string array isn't empty. This is the case in Circular.

In that case I think it's best to ignore the `get by alias` and `update by alias` functionality.

src/cmd/action.go
taimoor-at-opslevel: Added this change, also for trigger definitions and the delete() operations.

PR #167, opened 2023-09-13
print message when input coming from interactive tty

davidbloss:
Address case where command seems to hang - but really it's waiting for input from an interactive tty

### DIFF
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -1,32 +1,31 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 
-	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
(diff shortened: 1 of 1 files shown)


src/cmd/input.go
```diff
@@ -1,32 +1,31 @@
...
-		} else if dataFile == "." {
-			viper.SetConfigFile("./data.yaml")
-		} else {
-			viper.SetConfigFile(dataFile)
+	switch dataFile {
+	case ".":
+		viper.SetConfigFile("./data.yaml")
+	case "-":
+		if isStdInFromTerminal() {
+			fmt.Println("Reading input directly from command line...")
```
davidbloss, line 18: Send to stderr

PR #176, opened 2023-09-22
make `get tags` alway require a `key`, add `list tags` command to fetch all keys

taimoor-at-opslevel:
## Changelog

- [x] The `get tag KEY` command will no longer exit with a non-zero exit code when there are no tags that match `KEY`. Instead, it will just output an empty JSON array `[]`. This behaviour is more predictable - there being no matching tags is not a CLI/API error after all.
- [x] The `get tag` command now requires an argument, `KEY`. If the user wants to list all tags on an object, they have to use `list tags`
- [x] The `list tags` command is added and works the same as `get tag KEY`, except it just returns all tags without filtering for anything. 

### `get tag KEY`

Can find tags where `key == KEY`.

```
$ opslevel get tag --type team $PLATFORM newkey
[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwMTgwNTc3",
    "key": "newkey",
    "value": "newvalue"
  },
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwMTgwNTgx",
    "key": "newkey",
    "value": "secondvalue"
  }
]
```

Will not error out if no keys match `KEY`, will instead output empty JSON array.

```
$ opslevel get tag --type team $PLATFORM nonexistantkey
[]

```

### `list tags`

Should be able to output all tags.

```
$ opslevel list tags --type team $PLATFORM
[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwMTgwNTc3",
    "key": "newkey",
    "value": "newvalue"
  },
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwMTgwNTgx",
    "key": "newkey",
    "value": "secondvalue"
  },
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwMTgwNzMw",
    "key": "secondkey",
    "value": "rofl"
  }
]

```

If no tags exist, should output empty array.

```
~/repos/cli/src list-tags ❯ opslevel list tags --type team $PLATFORM                                                 Ruby 3.0.0 02:04:26 pm
[]

```

### DIFF
--- a/src/cmd/tag.go
+++ b/src/cmd/tag.go
@@ -135,17 +135,13 @@ opslevel get tag --type=Service ID|ALIAS KEY | jq
 		tags, err := GetTags(result)
 		cobra.CheckErr(err)
 
-		var output []opslevel.Tag
+		output := []opslevel.Tag{}
 		for _, tag := range tags.Nodes {
 			if tagKey == tag.Key {
 				output = append(output, tag)
 			}
 		}
 
-		// return empty JSON array instead of null
-		if len(output) == 0 {
-			output = make([]opslevel.Tag, 0)
-		}
 		common.PrettyPrint(output)
 	},
 }
--- a/src/cmd/tag.go
+++ b/src/cmd/tag.go
@@ -108,25 +108,20 @@ opslevel create tag --type=Team ID|ALIAS KEY VALUE
 }
 
 var getObjectTagCmd = &cobra.Command{
-	Use:   "tag --type=RESOURCE_TYPE RESOURCE_ID [KEY]",
-	Short: "Get values of tags on objects",
-	Long:  "Get values of tags on objects",
+	Use:   "tag --type=RESOURCE_TYPE RESOURCE_ID KEY",
+	Short: "Get value(s) of a tag on an object",
+	Long:  "Get value(s) of a tag on an object",
 	Example: `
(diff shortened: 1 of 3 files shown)


taimoor-at-opslevel:
I'd also like to add this gets rid of this weird behaviour (trying to get all tags when none exist results in this error message that shows an empty string, for some reason.)

```
opslevel get tag --type team platform
Error: tag with key '' not found on Team 'platform'
exit status 1
```

It is more clear if this just returns an empty array. What exactly is `key ''`?

src/cmd/tag.go
```diff
@@ -142,19 +137,51 @@ opslevel get tag --type=Team ID|ALIAS | jq 'from_entries'   # values of all keys
 
 		var output []opslevel.Tag
```
davidbloss, line 138: If `var output []opslevel.Tag` is substituted with
`output := []opslevel.Tag{}` 
then the `if len(output) == 0 { ... }` chunk below can be removed

davidbloss (approved):
I left one suggestion but it's not blocking.
Solid improvement! 🎉

rocktavious (approved):
LGTM

src/cmd/tag.go
taimoor-at-opslevel: nice! added this in latest commit.

Issue #159, opened 2023-08-25
As a CLI User, I want to see the last updated at for secrets when i list them out


rocktavious:
https://github.com/OpsLevel/cli/pull/157/files#r1306110432

PR #177, opened 2023-09-25
List secrets with last updated, input bugfix, create/update instructions update

taimoor-at-opslevel:
## Ticket

https://github.com/OpsLevel/cli/issues/159

## Changelog

- [x] Show last updated when listing secrets
- [x] Fix creating/updating secrets input reading and instructions

## Tophatting

Create a secret using interactive input:

```
~/repos/cli/src secrets-date !1 ❯ opslevel create secret --alias typed-secret-2 -f -                           х INT Ruby 3.0.0 12:51:39 pm
12:52PM INF Reading input directly from command line... Press CTRL+D to stop typing
owner:
  alias: "platform"
value: "made by typing directly"
&{Owner:{Id: Alias:platform} Value:made by typing directly}
Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY3
```

Create a secret using heredoc:

```
~/repos/cli/src secrets-date !1 ❯ cat << EOF | opslevel create secret --alias heredoc-secret-2 -f -               3s Ruby 3.0.0 12:50:35 pm
pipe heredoc> owner:
pipe heredoc>   alias: "platform"
pipe heredoc> value: "made by heredoc"
pipe heredoc> EOF
&{Owner:{Id: Alias:platform} Value:made by heredoc}
Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY2
```

Create a secret using a file:

```
~/repos/cli/src secrets-date !1 ❯ opslevel create secret --alias file-secret-2 -f ~/secret.yaml                  23s Ruby 3.0.0 12:50:25 pm
here: /Users/taimoorahmad/secret.yaml
&{Owner:{Id: Alias:platform} Value:this was made using dash f}
Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY1
```

Updated a secret using interactive input:

```
~/repos/cli/src secrets-date ❯ opslevel update secret Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY1 -f -      Ruby 3.0.0 12:58:32 pm
12:58PM INF Reading input directly from command line... Press CTRL+D to stop typing
owner:
  alias: "platform"
value: "updated this by typing"
Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY1
```

Finally, list secrets to see last updated time:

```
~/repos/cli/src secrets-date ❯ opslevel list secrets                                                           х INT Ruby 3.0.0 12:59:47 pm
ALIAS                  ID                                                    OWNER     UPDATED_AT
direct-secret          Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzYz  platform  2023-09-25 16:37:01.128483 +0000 UTC
file-secret-2          Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY1  platform  2023-09-25 16:59:04.213883 +0000 UTC
heredoc-secret         Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY0  platform  2023-09-25 16:46:25.341093 +0000 UTC
heredoc-secret-2       Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY2  platform  2023-09-25 16:51:34.104404 +0000 UTC
new-platform-secret    Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzYx  platform  2023-09-25 16:08:53.943065 +0000 UTC
stdin-platform-secret  Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzYy  platform  2023-09-25 16:11:28.950098 +0000 UTC
typed-secret-2         Z2lkOi8vb3BzbGV2ZWwvU2VjcmV0c1ZhdWx0czo6U2VjcmV0LzY3  platform  2023-09-25 16:52:25.989577 +0000 UTC
```

### DIFF
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/rs/zerolog/log"
@@ -12,25 +11,18 @@ import (
 var dataFile string
 
 func readInputConfig() {
-	fmt.Printf("dataFile is %s\n", dataFile)
-	viper.SetConfigType("yaml")
-
-	if dataFile == "." {
-		// TODO: validate file exists, if doesn't exist, check data.yml
-		fmt.Printf("using data.yaml\n")
+	switch dataFile {
+	case ".":
 		viper.SetConfigFile("./data.yaml")
-	} else if dataFile == "-" || dataFile == "" {
-		fmt.Printf("using stdin\n")
+	case "-":
 		if isStdInFromTerminal() {
 			log.Info().Msg("Reading input directly from command line...")
 		}
+		viper.SetConfigType("yaml")
 		viper.ReadConfig(os.Stdin)
-	} else {
-		// TODO: validate file exists
-		fmt.Printf("using %s\n", dataFile)
+	default:
 		viper.SetConfigFile(dataFile)
 	}
-
 	viper.ReadInConfig()
 }
 
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -12,19 +12,25 @@ import (
 var dataFile string
 
 func readInputConfig() {
-	viper.SetConfigType("yaml")
 	fmt.Printf("dataFile is %s\n", dataFile)
-	switch dataFile {
-	case ".":
+	viper.SetConfigType("yaml")
+
+	if dataFile == "." {
+		// TODO: validate file exists, if doesn't exist, check data.yml
+		fmt.Printf("using data.yaml\n")
 		viper.SetConfigFile("./data.yaml")
-	case "-":
+	} else if dataFile == "-" || dataFile == "" {
+		fmt.Printf("using stdin\n")
 		if isStdInFromTerminal() {
 			log.Info().Msg("Reading input directly from command line...")
 		}
 		viper.ReadConfig(os.Stdin)
-	default:
+	} else {
+		// TODO: validate file exists
+		fmt.Printf("using %s\n", dataFile)
 		viper.SetConfigFile(dataFile)
 	}
+
 	viper.ReadInConfig()
 }
 
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -18,6 +18,7 @@ func readInputConfig() {
 		viper.SetConfigFile("./data.yaml")
 	case "-":
 		if isStdInFromTerminal() {
+			// TODO: this can take up to half a second to output which interrupts the user's experience
 			log.Info().Msg("Reading input directly from command line... Press CTRL+D to stop typing")
 		}
 		viper.ReadConfig(os.Stdin)
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -11,19 +11,19 @@ import (
 var dataFile string
 
 func readInputConfig() {
+	viper.SetConfigType("yaml")
 	switch dataFile {
 	case ".":
+		// TODO: does this block ever actually ever run?
 		viper.SetConfigFile("./data.yaml")
 	case "-":
 		if isStdInFromTerminal() {
-			log.Info().Msg("Reading input directly from command line...")
+			log.Info().Msg("Reading input directly from command line... Press CTRL+D to stop typing")
 		}
-		viper.SetConfigType("yaml")
 		viper.ReadConfig(os.Stdin)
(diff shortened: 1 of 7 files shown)


src/cmd/input.go
```diff
@@ -11,14 +11,16 @@ import (
 var dataFile string
 
 func readInputConfig() {
+	viper.SetConfigType("yaml")
```
taimoor-at-opslevel, line 14: IIRC `-f filename.yaml` wasn't working until I moved this up here from outside of the `case "-"` block.

taimoor-at-opslevel: can't think of a reason for this to not be applied in certain conditions vs others, so it makes sense to move to the top.

src/cmd/input.go
```diff
@@ -11,14 +11,16 @@ import (
...
 	switch dataFile {
 	case ".":
+		// TODO: does this block ever actually ever run?
```
davidbloss, line 17: Somebody could be using this case.
Setting `-f` to `.` is called out in the command help docs for `opslevel create --help` ([create.go](https://github.com/OpsLevel/cli/blob/main/src/cmd/create.go#L17)) and `opslevel update --help` ([update.go](https://github.com/OpsLevel/cli/blob/main/src/cmd/update.go#L17)).

src/cmd/input.go
davidbloss: I think this is explicitly set to `yaml` below since `viper` has no way to infer the filetype from the stdin input.

[viper.SetConfigFile()](https://pkg.go.dev/github.com/spf13/viper@v1.16.0#SetConfigFile) defines the path, name and extension of the config file. Then `viper.ReadInConfig()` will know how to read it later.

If somebody is using a `.json` file I would assume viper would internally reset the configType below?

In any case, @rocktavious has mentioned that we should look into a better solution for reading user input in place of viper

src/cmd/input.go
```diff
@@ -11,14 +11,16 @@ import (
...
 		viper.SetConfigFile("./data.yaml")
 	case "-":
 		if isStdInFromTerminal() {
-			log.Info().Msg("Reading input directly from command line...")
+			// TODO: this can take up to half a second to output which interrupts the user's experience
```
davidbloss, line 21: 🤔 If this is working correctly users shouldn't see this unless they are manually typing in yaml formatted input. If that's not the case lmk and I can take a closer look

src/cmd/input.go
```diff
@@ -11,14 +11,16 @@ import (
...
+			log.Info().Msg("Reading input directly from command line... Press CTRL+D to stop typing")
```
davidbloss, line 22: Nice addition with `Press CTRL+D...`! 👍

davidbloss:
Overall looks good! Just left a few comments 👍

src/cmd/input.go
rocktavious: ya we need to fix this reading from input problems globally

rocktavious (approved):
LGTM

src/cmd/input.go
taimoor-at-opslevel: To clarify, this happens to me only when manually typing in the command. (The check `isStdInFromTerminal()` is fine)

What happens is: I start typing, then a second later the message pops up, I lose track of what I've typed, then `CTRL+C` and re-run the command and start typing again. It's a UX annoyance

src/cmd/input.go
taimoor-at-opslevel: I could not find a way to get it to trigger though, at least on bash/zsh. IIRC I ran into this problem before, passing `.` as an argument in CLI's is a bad idea 😅 

I don't think we should remove it immediately, but we should look into it because if this branch never gets called the `data.yaml` stuff is confusing.

PR #183, opened 2023-09-29
Db/implement taggable resources interface

davidbloss:
## Changelog

Update this tag logic to utilize opslevel-go's [add taggable resource interface](https://github.com/OpsLevel/opslevel-go/pull/250) changes.

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

```
GET
opslevel get tag --type=service Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS81Nzk3 "key-123"

[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy",
    "key": "key-123",
    "value": "val-123"
  }
]
```

```
# LIST
opslevel list tag --type=service Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS81Nzk3

[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy",
    "key": "key-123",
    "value": "val-123"
  }
]
```

```
# UPDATE
opslevel update tag Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy "test-key-123" "test-val-123"
o run ./... update tag Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy "test-key-123" "test-val-123"
{
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy",
    "key": "test-key-123",
    "value": "test-val-123"
}
```

```
# DELETE
opslevel delete tag Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy

deleted a tag: Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyNjYy
```

```
# CREATE
opslevel create tag --type=service Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS81Nzk3 "test-key-123" "test-value-123"
created tag for Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS81Nzk3: Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwOTEyODIy
```

### DIFF
--- a/src/cmd/tag.go
+++ b/src/cmd/tag.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 
 	"github.com/opslevel/cli/common"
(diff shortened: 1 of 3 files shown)


src/cmd/tag.go
```diff
@@ -120,28 +81,25 @@ opslevel get tag --type=Service ID|ALIAS KEY | jq
...
 		for _, tag := range tags.Nodes {
-			if tagKey == tag.Key {
-				output = append(output, tag)
+			if tag.Key == tagKey {
+				foundTag = tag
+				break
 			}
 		}
 
+		output := []opslevel.Tag{foundTag}
```
taimoor-at-opslevel, line 101: If there's no found tag, here's what I get on my machine:

```
~/repos/cli/src db/implement…es-interface ❯ opslevel get tag --type service $SERVICE ab                              Ruby 3.0.0 04:05:47 pm
[
  {
    "id": null,
    "key": "",
    "value": ""
  }
]

```

The output I expected was `[]` (empty JSON array).

taimoor-at-opslevel: There can be a case where a key has multiple values

```
~/repos/cli/src db/implement…es-interface ❯ opslevel list tag --type team platform                                   Ruby 3.0.0 04:09:04 pm
[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwNDI5NzAx",
    "key": "k",
    "value": "v"
  },
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwNDI5NzA2",
    "key": "k",
    "value": "v2"
  },
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwNDI5ODE5",
    "key": "k",
    "value": "v3"
  },
  ...
```

When a user tries to get `k`, the output is:

```
~/repos/cli/src db/implement…es-interface ❯ opslevel get tag --type team platform k                                  Ruby 3.0.0 04:09:14 pm
[
  {
    "id": "Z2lkOi8vb3BzbGV2ZWwvVGFnLzYwNDI5NzAx",
    "key": "k",
    "value": "v"
  }
]

```

But the correct output would be a JSON array of objects, where each object is a tag where `Tag.key` == `keyArg`

src/cmd/terraform.go
```diff
@@ -279,7 +279,12 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
 		if len(aliases) > 0 {
 			aliases = fmt.Sprintf("aliases = [\"%s\"]", aliases)
 		}
-		tags := flattenTags(service.Tags.Nodes)
+		serviceTags, err := service.GetTags(c, nil)
+		if err != nil {
```
taimoor-at-opslevel, line 283: I understand why we need a `// TODO` block, but I think `cobra.CheckErr(err)` would be better than the `if` block.

src/cmd/tag.go
```diff
@@ -160,18 +118,13 @@ opslevel list tag --type=Service ID|ALIAS
 		err := validateResourceTypeArg()
 		cobra.CheckErr(err)
 
-		resource := args[0]
+		resourceIdentifier := args[0]
 
-		var result any
-		if opslevel.IsID(resource) {
```
rocktavious, line 166: YES! 💪   I've been wanting to get this out of the CLI for a long time 🎉

taimoor-at-opslevel (changes requested):

src/cmd/tag.go
```diff
@@ -120,28 +81,25 @@ opslevel get tag --type=Service ID|ALIAS KEY | jq
...
-		output := []opslevel.Tag{}
+		tags, err := result.GetTags(client, nil)
+		cobra.CheckErr(err)
+		foundTag := opslevel.Tag{}
 		for _, tag := range tags.Nodes {
-			if tagKey == tag.Key {
-				output = append(output, tag)
+			if tag.Key == tagKey {
+				foundTag = tag
+				break
```
rocktavious, line 97: ahh ya @taimoor-at-opslevel is right we should not "break" here and instead build up a list of tags that match the key.

src/cmd/terraform.go
rocktavious: ya especially since we do that 8 lines up.  The end user for the most part is trying to dump out their terraform configuration so they'll see this error (it won't be in an automated system) so they will beable to react to the failure.

davidbloss: Ah, both good calls 👍  Putting this in

rocktavious (approved):

src/cmd/terraform.go
taimoor-at-opslevel: We can get rid of the `if err != nil` block altogether since we're using `cobra.CheckErr()`

taimoor-at-opslevel (changes requested):
Just requesting 1 change, otherwise LGTM.

taimoor-at-opslevel (approved):

PR #198, opened 2023-10-31
Db/deprecate groups

davidbloss:
## Issues

[#132](https://github.com/OpsLevel/team-platform/issues/132)

## Changelog

Deprecate `group` commands in favor of Team hierarchies.
`group create` and `group update` are no-ops. The other `group` commands direct devs to the docs and to use Teams.

- [X] List your changes here
- [X] Make a `changie` entry

### DIFF
--- a/src/cmd/group.go
+++ b/src/cmd/group.go
@@ -16,7 +16,7 @@ var createGroupCmd = &cobra.Command{
 	Use:        "group",
 	Short:      "Create a group (deprecated). Use teams instead.",
 	Long:       `Create a group (deprecated). Use teams instead.`,
-	Deprecated: `Please use Teams instead.`,
+	Deprecated: `Please use Teams instead. See https://docs.opslevel.com/docs/groups`,
 	Run: func(cmd *cobra.Command, args []string) {
 		err := errors.New("Groups are deprecated! Please use Teams instead.\nopslevel create team <args>")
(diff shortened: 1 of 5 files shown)


src/cmd/group.go
```diff
@@ -291,9 +271,10 @@ var deleteGroupCmd = &cobra.Command{
 }
 
 var importGroupsCmd = &cobra.Command{
-	Use:     "group",
-	Aliases: []string{"groups"},
-	Short:   "Imports groups from a CSV",
+	Use:        "group",
+	Aliases:    []string{"groups"},
+	Short:      "Imports groups from a CSV",
+	Deprecated: `Please convert all Groups into Teams. See https://docs.opslevel.com/docs/groups`,
```
rocktavious, line 277: Since you removed the create and update functionality - should we also remove the "import" functionality?

rocktavious (approved):
LGTM - just 1 question but otherwise good to ship.

taimoor-at-opslevel (approved):

src/cmd/group.go
davidbloss: I'm thinking that customers might use this to get a snapshot of their current Groups so that they can know what's left to convert to Teams

davidbloss: Removing per offline discussion

rocktavious (approved):

PR #203, opened 2023-11-13
drop func IsID() from common, exists in opslevel-go

davidbloss:
## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

```bash
# These commands continue to work
opslevel get repository <ID>
opslevel get service <ID>
opslevel get team platform
```


taimoor-at-opslevel (approved):
`code_quality++`

PR #206, opened 2023-11-27
Export Teams and team memberships to Terraform

taimoor-at-opslevel:
## Issues

After shipping https://github.com/OpsLevel/terraform-provider-opslevel/pull/144 some customers can be in a state where their terraform Team configs are not using the new `member { }` blocks. This results in Team Memberships not being able to reconcile without kicking all members out of the terraform managed teams.

[Also, the terraform export output is not compatible with the latest TF schema version.](https://registry.terraform.io/providers/OpsLevel/opslevel/0.8.13/docs/resources/team)

## Changelog

- [x] Change export team function
- [x] Minor changes: fix unnecessary line spaces, add line spaces between team resources to make it easier to read
- [x] Make a `changie` entry

## Tophatting

Before:

```
resource "opslevel_team" "platform" {
  name = "platform"
  manager_email = ""
  aliases = ["platform"]
  
}
resource "opslevel_team" "platform_2" {
  name = "Platform"
  manager_email = ""
  aliases = ["platform_2"]
  
}
```

After (places a line space between resources):

```
resource "opslevel_team" "platform" {
  aliases = ["platform"]
  name = "platform"
  member {
    email = "david+orange@opslevel.com"
    role = "contributor"
  }
  member {
    email = "kyle+orange@opslevel.com"
    role = "contributor"
  }
}

resource "opslevel_team" "platform_2" {
  aliases = ["platform_2"]
  name = "Platform"
  group = "platform"
  member {
    email = "taimoor+orange@opslevel.com"
    role = "contributor"
  }
  parent = ["engineering"]
}
```

### DIFF
--- a/src/cmd/terraform.go
+++ b/src/cmd/terraform.go
@@ -96,13 +96,13 @@ provider "opslevel" {
 
 	graphqlClient := getClientGQL()
 
-	// exportConstants(graphqlClient, constants)
-	// exportRepos(graphqlClient, repos, bash)
-	// exportServices(graphqlClient, bash, directory)
+	exportConstants(graphqlClient, constants)
+	exportRepos(graphqlClient, repos, bash)
+	exportServices(graphqlClient, bash, directory)
 	exportTeams(graphqlClient, teams, bash)
-	// exportFilters(graphqlClient, filters, bash)
-	// exportRubric(graphqlClient, rubric, bash)
-	// exportChecks(graphqlClient, bash, directory)
+	exportFilters(graphqlClient, filters, bash)
+	exportRubric(graphqlClient, rubric, bash)
+	exportChecks(graphqlClient, bash, directory)
 	fmt.Println("Complete!")
 }
 
--- a/src/cmd/terraform.go
+++ b/src/cmd/terraform.go
@@ -312,6 +312,7 @@ func exportTeams(c *opslevel.Client, config *os.File, shell *os.File) {
 
 	teamConfig := `resource "opslevel_team" "%s" {%s
 }
+
 `
 	resp, err := c.ListTeams(nil)
 	teams := resp.Nodes
--- a/src/cmd/terraform.go
+++ b/src/cmd/terraform.go
@@ -96,13 +96,13 @@ provider "opslevel" {
 
 	graphqlClient := getClientGQL()
 
-	exportConstants(graphqlClient, constants)
-	exportRepos(graphqlClient, repos, bash)
-	exportServices(graphqlClient, bash, directory)
+	// exportConstants(graphqlClient, constants)
+	// exportRepos(graphqlClient, repos, bash)
+	// exportServices(graphqlClient, bash, directory)
 	exportTeams(graphqlClient, teams, bash)
-	exportFilters(graphqlClient, filters, bash)
-	exportRubric(graphqlClient, rubric, bash)
-	exportChecks(graphqlClient, bash, directory)
+	// exportFilters(graphqlClient, filters, bash)
+	// exportRubric(graphqlClient, rubric, bash)
+	// exportChecks(graphqlClient, bash, directory)
 	fmt.Println("Complete!")
 }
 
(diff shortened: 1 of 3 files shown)


src/cmd/terraform.go
```diff
@@ -310,30 +310,49 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
...
+
 `
 	resp, err := c.ListTeams(nil)
 	teams := resp.Nodes
 	cobra.CheckErr(err)
 	for _, team := range teams {
+		teamBody := ""
+
 		aliases := flattenAliases(team.Aliases)
 		if len(aliases) > 0 {
```
davidbloss, line 324: As of [12 days ago](https://github.com/OpsLevel/terraform-provider-opslevel/pull/157/files#diff-c294bb8f5a3b817c0ee0a0e6054a395f6fde4d7cd4a9f8d55ff29700db4c4054R43) team `aliases` are required so this `len()` check should be removed.

src/cmd/terraform.go
```diff
@@ -310,30 +310,49 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
...
-			aliases = fmt.Sprintf("aliases = [\"%s\"]", aliases)
+			teamBody += fmt.Sprintf("\n  aliases = [\"%s\"]", aliases)
 		}
+		teamBody += fmt.Sprintf("\n  name = \"%s\"", team.Name)
+
+		if len(team.Group.Alias) > 0 {
```
davidbloss, line 329: Since groups are deprecated do you think this would be better left out?

If we keep it in we should create a follow up ticket to remove this when we officially remove [Group from the Team struct](https://github.com/OpsLevel/opslevel-go/blob/main/team.go#L53)

src/cmd/terraform.go
```diff
@@ -310,30 +310,49 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
...
-  %s
+	teamConfig := `resource "opslevel_team" "%s" {%s
 }
+
 `
 	resp, err := c.ListTeams(nil)
 	teams := resp.Nodes
 	cobra.CheckErr(err)
 	for _, team := range teams {
+		teamBody := ""
```
davidbloss, line 321: Can we please introduce `strings.Builder` here?
We can set `teamBody := strings.Builder{}` just above this for loop then:
1. sb.WriteString("strings below")
2. sb.String()  // to return the whole string to `config.WriteString(templateConfig()` below
3. sb.Reset() // to reset the underlying buffer at the end of each loop iteration

Running this command is surprisingly slow and this would be a small step in a more optimized direction IMO

src/cmd/terraform.go
```diff
@@ -310,30 +310,49 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
...
+		membersOutput := ""
+		for _, member := range team.Memberships.Nodes {
+			memberConfig := `
+  member {
+    email = "%s"
+    role = "%s"
+  }`
+			membersOutput += fmt.Sprintf(memberConfig, member.User.Email, member.Role)
+		}
+		if len(membersOutput) > 0 {
```
davidbloss, line 341: This `if len()` check is safe to remove since appending an empty string doesn't really change teamBody

davidbloss: Actually where `membersOutput` starts down to here seems like a good candidate for a small function. Lmk if you would like to discuss :) 
Something loosely like `membersOutput := getMembershipsAsTerraformConfig(team.Memberships.Nodes)`

src/cmd/terraform.go
taimoor-at-opslevel: I think it should be kept in up until we remove the field entirely from terraform. Right now the field exists, but is deprecated. There may be some customers who have not migrated away from groups and would appreciate having this in the export. 

So the follow-up ticket should be the way to go.

src/cmd/terraform.go
taimoor-at-opslevel: Implemented these suggestions

src/cmd/terraform.go
taimoor-at-opslevel: Added!

src/cmd/terraform.go
```diff
@@ -307,35 +307,55 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
 	}
 }
 
+func getMembershipsAsTerraformConfig(members []opslevel.TeamMembership) string {
+	membersBody := strings.Builder{}
+	for _, member := range members {
+		memberConfig := `
```
davidbloss, line 313: Can `memberConfig` be moved out of the for loop?
Overall this looks good! 👍

src/cmd/terraform.go
```diff
@@ -307,35 +307,55 @@ func exportServices(c *opslevel.Client, shell *os.File, directory string) {
...
-  %s
-  %s
+	teamConfig := `resource "opslevel_team" "%s" {%s
 }
+
 `
 	resp, err := c.ListTeams(nil)
 	teams := resp.Nodes
 	cobra.CheckErr(err)
+	teamBody := strings.Builder{}
```
davidbloss, line 333: I'm really excited about this! 🙌

src/cmd/terraform.go
taimoor-at-opslevel: Moved!

davidbloss (approved):
LGTM! 🎉

PR #209, opened 2023-12-08
consolidate cli input read funcs into readResourceInput

davidbloss:
## Changelog

- Consolidate similar readThing() funcs into `readResourceInput()` at [src/cmd/input.go](https://github.com/OpsLevel/cli/compare/db/unify-read-cli-input-to-readResourceInput?expand=1#diff-03faa53a7613111d9e5a128a4cd786a590a5278c48fbde3d50558ffa4e8e902e)
```go
func readResourceInput[T any]() (*T, error) {
	readInputConfig()
	var evt *T
	viper.Unmarshal(&evt)
	if err := defaults.Set(evt); err != nil {
		return nil, err
	}
	return evt, nil
}
```

- A change similar to below is applied in multiple places:
```diff
- input, err := readWebhookActionCreateInput()
+ input, err := readResourceInput[opslevel.CustomActionsWebhookActionCreateInput]()
```
- The defunct readThing() funcs were then removed

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

Manual test commands: 
```bash
# In orange
TEAM_ID=$(opslevel get team tf_baz | jq .Id | tr -d '"')
opslevel update team $TEAM_ID -f -
responsibilities: "work on many things"
^D

opslevel get team tf_baz | jq .Responsibilities
```

### DIFF
--- a/.changes/unreleased/Feature-20231208-102655.yaml
+++ b/.changes/unreleased/Feature-20231208-102655.yaml
@@ -1,3 +1,0 @@
-kind: Feature
-body: consolidate cli input read funcs into readResourceInput
-time: 2023-12-08T10:26:55.311708-06:00
--- a/.changes/unreleased/Feature-20231208-102655.yaml
+++ b/.changes/unreleased/Feature-20231208-102655.yaml
@@ -1,0 +1,3 @@
+kind: Feature
+body: consolidate cli input read funcs into readResourceInput
+time: 2023-12-08T10:26:55.311708-06:00
--- a/src/cmd/input.go
+++ b/src/cmd/input.go
@@ -1,12 +1,13 @@
 package cmd
 
 import (
+	"bytes"
 	"os"
 
 	"github.com/rs/zerolog/log"
 
-	"github.com/creasty/defaults"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
 )
 
 var dataFile string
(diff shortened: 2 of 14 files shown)


.changes/unreleased/Feature-20231208-102655.yaml
taimoor-at-opslevel, line 1: I don't think this is necessary, because the customer deliverable in CLI is the binary itself and not the code unlike in opslevel-go.

taimoor-at-opslevel (approved):
Excellent!

.changes/unreleased/Feature-20231208-102655.yaml
rocktavious: Agree

davidbloss: Yeah, I was on the fence about this. I'll drop this since nothing customer facing has changed.

src/cmd/input.go
```diff
@@ -28,6 +29,16 @@ func readInputConfig() {
 	viper.ReadInConfig()
 }
 
+func readResourceInput[T any]() (*T, error) {
+	readInputConfig()
+	var evt *T
```
rocktavious, line 34: This should start out as a `T` and then be returned at the return as `*evt`

davidbloss: Good call 👍

src/cmd/input.go
```diff
@@ -28,6 +29,16 @@ func readInputConfig() {
...
+	viper.Unmarshal(&evt)
```
rocktavious, line 35: We shouldn't use viper here - it conflates with the options vars and could cause unexpected problems if a CLI flag overlaps with an input type field in name.  We should use `yaml.Unmarshal` by reading the input file using `os.ReadFile`.

This has been on my backburner to fix

rocktavious (changes requested):
2 changes

rocktavious (approved):

PR #210, opened 2023-12-08
make Dependabot PRs trigger changie generating Github workflow

davidbloss:
## Changelog

Use [reusable workflow from our actions repo](https://github.com/OpsLevel/actions/blob/main/.github/workflows/dependabot_pr_labels.yml) to trigger adding a changelog to Dependabot PRs with changie.

Note: the job in this workflow runs only if dependabot opens a PR. If we want to also keep a manual way of triggering the chanige job lmk and I can add some extra "triggered by" stuff [there](https://github.com/OpsLevel/opslevel-go/blob/main/.github/workflows/changie-gen.yml)

- [X] List your changes here
- [ ] Make a `changie` entry, N/A CI only


taimoor-at-opslevel (approved):

taimoor-at-opslevel:
@davidbloss will this end up changing the currently open dependabot PR's we have open to create changie entries?

PR #214, opened 2023-12-11
Db/manage property definitions

davidbloss:
## Changelog

- Add support for PropertyDefinitions
- Add `gen` command that generates example yaml based on given structs
  - Only `opslevel gen property-definition` added here

- [X] List your changes here
- [X] Make a `changie` entry

Extra changes actually match up with what's in main. Artifacts of rebasing

## Tophatting

```bash
# Use generate command to create yaml requirements for `PropertyDefinitionInput{}`
opslevel gen property-definition > test-prop.yaml

# Create PropertyDefinition with example data, save id in env variable
PROP_DEF_ID=$(opslevel create property-definition -f test-prop.yaml)

# Get new PropertyDefinition
opslevel get property-definition $PROP_DEF_ID

# List all PropertyDefinitions
opslevel list property-definitions

# Delete new PropertyDefinition
opslevel delete property-definition $PROP_DEF_ID
```


src/cmd/property.go
```diff
@@ -0,0 +1,116 @@
...
+	Short:      "Get details about a property definition",
+	Long:       `Get details about a property definition`,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
+	Run: func(cmd *cobra.Command, args []string) {
+		identifier := args[0]
+		result, err := getClientGQL().GetPropertyDefinition(identifier)
+		cobra.CheckErr(err)
+
+		common.WasFound(result.Id == "", identifier)
```
taimoor-at-opslevel, line 44: `GetPropertyDefinition` will return an error anyways if the resource isn't found, so this line is unnecessary. This is the case with a LOT of code in tools that calls `GetSomeResource()`.

src/cmd/property.go
```diff
@@ -0,0 +1,116 @@
...
+	Use:        "property-definition ID",
+	Short:      "Delete a property definitions",
+	Long:       "Delete a property definitions",
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
+	Run: func(cmd *cobra.Command, args []string) {
+		propertyDefinitionId := args[0]
+		err := getClientGQL().DeletePropertyDefinition(propertyDefinitionId)
+		cobra.CheckErr(err)
+		fmt.Printf("deleted '%s' property definition\n", propertyDefinitionId)
```
taimoor-at-opslevel, line 106:
```suggestion
		fmt.Printf("deleted property definition '%s'\n", propertyDefinitionId)
```

taimoor-at-opslevel (changes requested):
Couple change requests

src/cmd/property.go
```diff
@@ -0,0 +1,116 @@
...
+			Name:   "Example Property Definition",
+			Schema: opslevel.JSONString(schema),
+		})
+	if err != nil {
+		cobra.CheckErr(err)
+	}
+	return strings.TrimSpace(examplePropertyDefinition)
+}
+
+var listPropertyDefinitionsCmd = &cobra.Command{
```
taimoor-at-opslevel, line 75: This function should have an alias like other `list` commands, so I can do both

```
opslevel list property-definitions
opslevel list property-definition
```

src/cmd/property.go
davidbloss: Good call 👍

taimoor-at-opslevel (approved):

src/cmd/property.go
```diff
@@ -0,0 +1,116 @@
...
+	Use:   "property-definition",
+	Short: "Generate example yaml of a Property Definition",
+	Long:  `Generate example yaml of a Property Definition`,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println(getExamplePropertyDefinitionYaml())
+	},
+}
+
+func getExamplePropertyDefinitionYaml() string {
+	schema := `{"$schema":"https://json-schema.org/draft/2020-12/schema", "type": "boolean"}`
```
rocktavious, line 62: we should use a more featureful example - can you reachout to foundations and ask what examples/templates they are putting in the UI and put one of the more complex ones here?

rocktavious (approved):

src/cmd/property.go
davidbloss: Definitely! The more feature rich example will live in opslevel-go

PR #219, opened 2023-12-18
add dependabot gitsubmodule check for end of week

davidbloss:
## Changelog

Add [gitsubmodule check](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) to `dependabot` config as an extra reminder to update submodules at the end of the week. This can help ensure the cli stays up to date with the latest opslevel-go code

Setting `branch = main` in `.gitmodules` lets `task setup` use `git submodule update --remote` to always get the latest commit from `main`

- [X] List your changes here
- [ ] Make a `changie` entry, N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,13 @@ updates:
     directory: "src/" # Location of package manifests
     schedule:
       interval: "weekly"
+  - package-ecosystem: "gitsubmodule"
+    directory: "src/"
+    schedule:
+      interval: "weekly"
+      day: "friday"
+      time: "10:00"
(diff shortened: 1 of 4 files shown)


taimoor-at-opslevel (approved):
Is there any reason to not make this daily? 🤔

davidbloss:
> Is there any reason to not make this daily? 🤔

My take is this should be an extra safety net around the time we release the code. But a primary (automatic) solution is what we need, and I updated `task setup` to automatically pull the latest from `opslevel-go`'s main branch.

PR #222, opened 2023-12-18
use tasks to stay on opslevel-go latest release

davidbloss:
## Changelog

Today we had a hiccup in releasing the this repo and these changes aim to make future releases more reliable.

Adding `task: has-latest-opslevel-go` as a command to `task lint` now checks if the version of opslevel-go known to the cli matches the latest opslevel-go release version. `task lint` itself has been added to the test workflow.

- [X] List your changes here
- [ ] Make a `changie` entry, N/A taskfile change


Taskfile.yml
```diff
@@ -3,13 +3,15 @@
 version: '3'
 
 env:
+  OPSLEVEL_GO_PKG: "github.com/opslevel/opslevel-go/v2023"
```
davidbloss, line 6: We will need to manually update this for `v2024`

Taskfile.yml
```diff
@@ -61,6 +63,27 @@ tasks:
...
+    dir: "{{.SRC_DIR}}"
+    silent: true
+    vars:
+      GO_WORK: "go.work"
+      TMP_GO_WORK: "tmp_go.work"
+      LATEST_OPSLEVEL_GO_VERSION:
+        sh: go list -u -m -versions github.com/opslevel/opslevel-go/v2023 | awk '{print $NF}'
+    cmds:
+      - if [[ -f {{.GO_WORK}} ]]; then mv {{.GO_WORK}} {{.TMP_GO_WORK}}; fi
+      - defer: if [[ -f {{.TMP_GO_WORK}} ]]; then mv {{.TMP_GO_WORK}} {{.GO_WORK}}; fi
```
davidbloss, line 77: For this task, `go.work` is briefly moved so `go list` finds the `.Version` in `go.mod` instead of a `.Path` to the submodule.

.github/workflows/tests.yml
```diff
@@ -25,10 +25,12 @@ jobs:
         with:
           version: 3.x
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Run Tests
+      - name: Lint code
```
taimoor-at-opslevel, line 28: This isn't clear to me, why does the workflow have to have a linting step? I thought when the workflow runs it should expect the code to already be linted. 

To me it would make more sense for the `Test code` step to also do `task setup` and for this `Lint code` step to not be added.

Taskfile.yml
```diff
@@ -3,13 +3,15 @@
...
 env:
+  OPSLEVEL_GO_PKG: "github.com/opslevel/opslevel-go/v2023"
   SRC_DIR: "{{.TASKFILE_DIR}}/src"
 
 tasks:
   lint:
     desc: Formatting and linting
     dir: "{{.SRC_DIR}}"
     cmds:
+      - task: has-latest-opslevel-go
```
taimoor-at-opslevel, line 14: Along with the other comment, I think it's unusual that `task lint` and `task lintfix` include a step to grab the latest opslevel-go. I don't think that this would fall under formatting or linting.

taimoor-at-opslevel (changes requested):

.github/workflows/tests.yml
davidbloss: `task lint` verifies that code has been linted - which is lacking from this CI

`task test` only checks a fraction of the code in this repo whereas `task lint` reviews all the go code.
Also, testing doesn't catch a long list of things that could be problematic but don't actively cause problems.

For example
```go
// golangci-lint points out that "Error return value of `json.Unmarshal` is not checked"
json.Unmarshal([]byte(data), &result)

// Which would result in us updating the above to this below
if err := json.Unmarshal([]byte(data), &result); err != nil {
	panic(err)
}
```
Both would pass in our tests, but one is much easier to debug.

taimoor-at-opslevel (approved):
There's a merge conflict in go.sum as I am typing this but the new version looks better

PR #224, opened 2023-12-21
add create,get,list,delete PropertyDefinitions capabilities

davidbloss:
## Changelog

Manage PropertyDefinitions with the cli

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

```bash
opslevel create property-definition -f <(opslevel example property-definition)
# ID returned
opslevel get property-definition <ID>
opslevel list property-definitions
opslevel delete property-definition <ID>
```

### DIFF
--- a/src/cmd/property.go
+++ b/src/cmd/property.go
@@ -1,102 +1,132 @@
 package cmd
 
-// import (
-// 	"encoding/json"
-// 	"fmt"
+import (
+	"encoding/json"
+	"fmt"
 
-// 	"github.com/opslevel/opslevel-go/v2023"
+	"github.com/opslevel/opslevel-go/v2023"
 
-// 	"github.com/opslevel/cli/common"
+	"github.com/opslevel/cli/common"
 
-// 	"github.com/spf13/cobra"
-// )
+	"github.com/spf13/cobra"
+)
 
-// var examplePropertyDefinitionCmd = &cobra.Command{
-// 	Use:   "property-definition",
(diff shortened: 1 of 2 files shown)


src/cmd/property.go
```diff
@@ -1,102 +1,132 @@
...
+	schema, ok := data["schema"].(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("schema is required and must be a JSON object")
+	}
+	propDefInput := opslevel.PropertyDefinitionInput{
+		Name:   name,
+		Schema: opslevel.JSON(schema),
+	}
+
+	if description, ok := data["description"].(string); !ok {
```
taimoor-at-opslevel, line 61:
```suggestion
	if description, ok := data["description"].(string); ok {
```

Noticed that `description` will not be read because of `!ok`.

src/cmd/property.go
```diff
@@ -1,102 +1,132 @@
...
+		propDefInput.Description = description
+	}
+	if propertyDisplayStatus, ok := data["propertyDisplayStatus"].(string); !ok {
```
taimoor-at-opslevel, line 64:
```suggestion
	if propertyDisplayStatus, ok := data["propertyDisplayStatus"].(string); ok {
```

taimoor-at-opslevel (changes requested):
Noticed a bug with reading the optional fields, otherwise LGTM/WOMM

taimoor-at-opslevel (approved):

PR #239, opened 2024-01-20
Increase convenience by adding more aliases for commands

taimoorgit:
## Issues

This is work I had in the CLI gen branch that I want to merge separately.

## Changelog

- [x] Add aliases for almost all commands. Each list command should support singular and plural versions of any aliases.
- [x] Add `ls`/`rm` as aliases to list/delete
- [x] Make a `changie` entry


src/cmd/team.go
```diff
@@ -12,17 +12,19 @@ import (
...
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println(getExample[opslevel.TeamCreateInput]())
 	},
 }
 
 var createTeamCmd = &cobra.Command{
-	Use:   "team NAME",
-	Short: "Create a team",
+	Use:     "team NAME",
+	Aliases: common.GetAliases("Team"),
```
davidbloss, line 26: I've just started looking at these. The help docs for this (`opslevel create team --help`) show me:
```bash
Usage:
  opslevel create team NAME [flags]

Aliases:
  team, team, teams
```
Which is actually pretty confusing.
1. `team` is the default already and listed twice under `Aliases:`
2. `teams` suggests that I can create multiple teams with this one command. But the usage says `Short: "Create a team"`

src/cmd/team.go
```diff
@@ -134,8 +136,9 @@ opslevel create contact --type=email my-team team@example.com "Mailing List"`,
 }
 
 var updateTeamCmd = &cobra.Command{
-	Use:   "team {ID|ALIAS}",
-	Short: "Update a team",
+	Use:     "team {ID|ALIAS}",
+	Aliases: common.GetAliases("Team"),
```
davidbloss, line 140: Same [as above](https://github.com/OpsLevel/cli/pull/239/files#r1462084462)

src/cmd/team.go
```diff
@@ -182,7 +186,7 @@ func GetTeam(key string) (*opslevel.Team, error) {
 
 var listTeamCmd = &cobra.Command{
 	Use:     "team",
-	Aliases: []string{"teams"},
+	Aliases: common.GetAliases("Team"),
```
davidbloss, line 189: Here specifically, it seems like it makes more sense to leave `Aliases: []string{"teams"},` as is

src/common/grammar.go
```diff
@@ -0,0 +1,62 @@
+package common
```
davidbloss, line 1: `grammar.go`is an overloaded name, IMO. I opened this expecting some parsing constructs - [example grammar.go](https://github.com/golang/go/blob/master/src/mime/grammar.go)

Since this logic is only used in `alias.go`, it might make sense to move it all over there.

src/common/grammar.go
```diff
@@ -0,0 +1,62 @@
...
+		i++
+	}
+	return res
+}
+
+func KebabCase(s string) string {
+	return strings.ReplaceAll(SnakeCase(s), "_", "-")
+}
+
+func IndefArticle(s string) string {
```
davidbloss, line 40: ❔ `IndefArticle` is unused

src/common/grammar.go
```diff
@@ -0,0 +1,62 @@
...
+		return "an"
+	}
+	return "a"
+}
+
+func IsPlural(s string) bool {
+	return strings.HasSuffix(s, "s")
+}
+
+func Pluralize(s string) string {
```
davidbloss, line 53: Alternative approach. Instead of Pluralizing some aliases with potential side effects, can we just call `GetAliases()` with multiple args?

`common.GetAliases("Deploy", "Deploys")`

src/common/alias.go
```diff
@@ -0,0 +1,41 @@
...
+
+// GetAliases will return a slice of variations of a PascalCase string like the
+// lowercase, snake_case, kebab-case, singular and plural versions
+func GetAliases(PascalCase string) []string {
+	if _, ok := hardcodedAliases[PascalCase]; !ok {
+		hardcodedAliases[PascalCase] = []string{}
+	}
+	addAlias(PascalCase, strings.ToLower(PascalCase))
+	addAlias(PascalCase, SnakeCase(PascalCase))
+	addAlias(PascalCase, KebabCase(PascalCase))
```
davidbloss, line 35: How do you feel about moving
```go
if _, ok := hardcodedAliases[PascalCase]; !ok {
	hardcodedAliases[PascalCase] = []string{}
}
addAlias(PascalCase, strings.ToLower(PascalCase))
addAlias(PascalCase, SnakeCase(PascalCase))
addAlias(PascalCase, KebabCase(PascalCase))
```
into `addAlias()` above?

Just feels like a lot of mutations living inside a `Get` function

src/cmd/team.go
taimoorgit: @davidbloss this has been fixed.

1. Aliases are not shown unnecessarily in the help menu anymore
2. Example, create, update, delete, assign, unassign don't support plural nouns because it can cause confusion
3. Only list, import support plural nouns

Create team (no aliases) looks normal again:

```
$ op create team -h
Create a team

Usage:
  opslevel create team NAME [flags]
```

Here's create system (has alias) and list system:

```
$ op create system -h
Create a system

Usage:
  opslevel create system [flags]

Aliases:
  system, sys

$ op list system -h
Lists the systems

Usage:
  opslevel list system [flags]

Aliases:
  system, systems, sys
```

taimoorgit:
@rocktavious and @davidbloss PR has been reworked to be way simpler. [Some output is shown here.](https://github.com/OpsLevel/cli/pull/239#discussion_r1462227477)

src/cmd/list.go
```diff
@@ -8,9 +8,10 @@ import (
 var listOutputType string
 
 var listCmd = &cobra.Command{
-	Use:   "list",
-	Short: "List all resources in OpsLevel",
-	Long:  "List all resources in OpsLevel",
+	Use:     "list",
+	Aliases: []string{"ls"},
```
rocktavious, line 12: neat!

src/cmd/property.go
```diff
@@ -221,8 +229,8 @@ var getPropertyDefinitionCmd = &cobra.Command{
 
 var listPropertyDefinitionsCmd = &cobra.Command{
 	Use:     "property-definition",
+	Aliases: []string{"property-definitions", "propertydefinition", "propertydefinitions", "pd", "pds"},
```
rocktavious, line 232: `propdef` should we add this too?

src/cmd/trigger_definition.go
```diff
@@ -11,18 +11,20 @@ import (
 )
 
 var exampleTriggerDefinitionCmd = &cobra.Command{
-	Use:   "trigger-definition",
-	Short: "Example Scorecard",
-	Long:  `Example Scorecard`,
+	Use:     "trigger-definition",
+	Aliases: []string{"triggerdefinition", "td"},
+	Short:   "Example Scorecard",
+	Long:    `Example Scorecard`,
```
rocktavious, line 17: these are not correct descriptions

rocktavious (approved):
Overall this is MUCH easier to grok.  There was 1 problem i noticed in surrounding code but all in all this is a much better change. 🎉

src/cmd/property.go
taimoorgit: Added `propdef` and `trigdef`

PR #241, opened 2024-01-21
grammar fixes, quote vars in bash, fix README links

taimoorgit:


ci/deploy-deb.sh
```diff
@@ -1,21 +1,23 @@
 #!/bin/bash
 
+set -e
```
taimoorgit, line 3: Added so that the CI can just exit if an error is encountered, preventing the script from trying to continue to run.

ci/deploy-deb.sh
```diff
@@ -1,21 +1,23 @@
...
+
 DEBIAN_RELEASES=$(debian-distro-info --supported)
 UBUNTU_RELEASES=$(ubuntu-distro-info --supported-esm)
 
 mkdir -p cli-repo/deb
 cd cli-repo/deb
 
-for release in ${DEBIAN_RELEASES[@]} ${UBUNTU_RELEASES[@]}; do
+for release in "${DEBIAN_RELEASES[@]}" "${UBUNTU_RELEASES[@]}"; do
```
davidbloss, line 11: This raises the question.
_What bash/shell linters are we using for this project?_

I know that my shell linter config is different than Kyle's config, and our "format on save" features are overwriting each others quote / curly braces defaults. This is a small annoyance that we should reconcile as a team.

davidbloss (approved):
LGTM! 🎉

PR #253, opened 2024-02-27
Re-enable linters in CLI

taimoorgit:
Depends on https://github.com/OpsLevel/cli/pull/241

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,9 +1,16 @@
 run:
   timeout: 2m
-linters:
-  disable:
-      # 1st -> remove staticcheck and fix lint issues
-    - staticcheck
-      # 2nd -> remove errcheck and ineffassign
-    - errcheck
-    - ineffassign
+linters-settings:
+  errcheck:
+    exclude-functions:
+      - (*encoding/csv.Writer).Write
+      - (*github.com/spf13/cobra.Command).MarkFlagRequired
+      - (*github.com/spf13/pflag.PFlagSet).GetString
+      - (*os.File).WriteString
+      - github.com/spf13/viper.BindEnv
+      - github.com/spf13/viper.BindPFlags
+      - github.com/spf13/viper.ReadConfig
+      - github.com/spf13/viper.ReadInConfig
+      - github.com/spf13/viper.Unmarshal
+  staticcheck:
+    checks: ["all", "-SA5001"]
--- a/src/cmd/graphql.go
+++ b/src/cmd/graphql.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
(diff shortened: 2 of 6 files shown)


.golangci.yml
```diff
@@ -1,9 +1,16 @@
...
+      - (*github.com/spf13/cobra.Command).MarkFlagRequired
+      - (*github.com/spf13/pflag.PFlagSet).GetString
+      - (*os.File).WriteString
+      - github.com/spf13/viper.BindEnv
+      - github.com/spf13/viper.BindPFlags
+      - github.com/spf13/viper.ReadConfig
+      - github.com/spf13/viper.ReadInConfig
+      - github.com/spf13/viper.Unmarshal
+  staticcheck:
+    checks: ["all", "-SA5001"]
```
taimoorgit, line 16: Excluded https://staticcheck.io/docs/checks/#SA5001

This prevents us from being able to do `defer thing.Close()` in a function because it forces us to check the error.

src/cmd/graphql.go
```diff
@@ -9,9 +9,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
```
davidbloss, line 12: `"github.com/pkg/errors"` is a [publicly archived](https://github.com/pkg/errors) project with a release date of 4 years ago.
If we really want the slick error wrapping feature, let's just pull the good stuff out of that repo and add it inside of `handleErr()`.
That also gets us one step closer to a "unified error message" if that's a long term goal

.golangci.yml
```diff
@@ -1,9 +1,16 @@
...
   timeout: 2m
-linters:
-  disable:
-      # 1st -> remove staticcheck and fix lint issues
-    - staticcheck
-      # 2nd -> remove errcheck and ineffassign
-    - errcheck
-    - ineffassign
+linters-settings:
+  errcheck:
```
davidbloss, line 4: Great to see! 🚀

davidbloss (changes requested):
If we want to beef up error messages that's all good, let's just do that without adding an archived dependency

src/cmd/graphql.go
```diff
@@ -169,13 +172,6 @@ func init() {
 	graphqlCmd.Flags().StringArrayP("field", "f", nil, "Add a variable in `key=value` format")
 }
 
-func handleErr(msg string, err error) {
```
davidbloss, line 172: We can add `cobra.CheckErr()` once here

taimoorgit: @davidbloss restored that function and removed all mentions of `errors` package. Now using `fmt.Errorf` everywhere to make errors, which is the idiomatic way.

davidbloss (approved):
LGTM! 🎉

PR #255, opened 2024-03-05
refactor input functions to make them testable

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/264

## Changelog

- [x] Divide `readResourceInput` into 3 functions: one to read input, one to apply input, one to do both
- [x] Change input function signatures to allow injecting test data by using variadic arguments
- [x] Add tests for input functions


src/cmd/input.go
```diff
@@ -30,34 +34,108 @@ func readInputConfig() {
...
+	// TODO: multiline JSON is not supported as a value input because handling whitespace and escape characters is nontrivial
+	jsonString, err := opslevel.NewJSONInput(m["value"])
+	if err != nil {
+		return err
+	}
+	m["value"] = jsonString
+	return nil
+}
+
+func ReadResourceHandleJSONFields[T any](input []byte) (*T, error) {
```
rocktavious, line 92: It looks like this is ONLY used in test files - this shouldn't be part of the codebase then it should be a utility function in the testing code.

rocktavious: Same with its helper functions `handleJSONString` and `handleJSONSchema`

src/cmd/input.go
```diff
@@ -30,34 +34,108 @@ func readInputConfig() {
...
+
+	var finalInput T
+	err = mapstructure.Decode(toMap, &finalInput)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding map as type %T: %w", finalInput, err)
+	}
+	return &finalInput, nil
+}
+
+func ReadResourceInput[T any](input []byte) (*T, error) {
```
rocktavious, line 128: The only reason `input` is here is because of the tests - we should refactor this so it doesn't take input (because the only thing i should be doing is "reading input"

The function that takes an input argument should be a helper function in the test code.

src/cmd/input.go
taimoorgit: @rocktavious this is used in the properties PR that follows this: https://github.com/OpsLevel/cli/pull/257

rocktavious: In fact after reading the whole code - i don't see where this function is used - so why are we testing it?

src/cmd/input.go
```diff
@@ -30,34 +34,108 @@ func readInputConfig() {
...
+		}
+	}
+
+	m, err := ReadResource[map[string]any](input)
+	if err != nil {
+		return nil, fmt.Errorf("error creating map from input: %w", err)
+	}
+	toMap := *m
+
+	if _, ok := toMap["schema"]; ok {
```
rocktavious, line 107: this feels wrong.  This feels like we need to implement custom yaml mashaller functions for `PropertyInput` and `PropertyDefinitionInput` ?

src/cmd/input.go
taimoorgit: @rocktavious being able to pass in `input` can serve a purpose outside of tests as well, for example by being to do `readInput()` to get input in bytes that can be used to get the raw input from the user before calling `ReadResourceInput()`.

src/cmd/input.go
taimoorgit: @rocktavious I wanted to add an abstraction in case we would start adding more resource types that use JSON fields outside of properties, but I have now reversed this.

src/cmd/input.go
rocktavious: This add complexity for a "what if" that we are currently not doing in our code outside of testing nor would likely do IMO.

Is there any reason that the tests cannot just use `ReadResource[T](input)` ?

src/cmd/input.go
rocktavious: I feel like the abstraction is more complex and unexpected then having custom yaml marshallars

src/cmd/input.go
taimoorgit: @rocktavious will address that in the follow-up PR that handles `JSONSchema`/`JSONString`

src/cmd/input.go
taimoorgit: @rocktavious I have changed the `ReadResourceInput` signature to use a variadic argument for the mocked inputs. Is that OK? It means that we don't need to `(nil)` every time.

src/cmd/input.go
taimoorgit: @rocktavious if we remove support for JSON, we could replace this code with only custom YAML marshalers which would be much cleaner. Supporting both YAML and JSON feels hacky. 

Also, while you can read in JSON objects, reading in JSON values (number, string, boolean, list) is next to impossible without us writing another 100+ lines of code. 

How do you feel about only supporting YAML?

src/cmd/input.go
rocktavious: thats definitely better - i still don't like it because now you are abusing variadic - especially because you  don't handle multiple arguments - but we can move on since this gets what i'm after it just leaves tech debt in our codebase.

src/cmd/input.go
rocktavious: https://github.com/OpsLevel/cli/pull/255#discussion_r1514876000

PR #257, opened 2024-03-05
add support for PropertyDisplayStatus and Locked to CLI | add property tests

taimoorgit:
## Issues

https://github.com/OpsLevel/team-platform/issues/264

Depends on: https://github.com/OpsLevel/cli/pull/255

## Changelog

TODO

## Tophatting

TODO: YAML only

- [ ] Can create a simple value property definition
- [ ] Can assign a simple value property
- [ ] Can update it into an object using YAML
- [ ] Can assign an object property
- [ ] Can update it into a list using YAML
- [ ] Can assign a list property


.changes/unreleased/Feature-20240305-194642.yaml
```diff
@@ -0,0 +1,3 @@
+kind: Feature
```
rocktavious, line 1: when you find yourself adding more then 1 changie - you are doing too much work in 1 PR.  It would be much better to have done these things isolated in smaller PRs that are dedicated to making 1 change.

src/cmd/property.go
```diff
@@ -15,18 +16,63 @@ var examplePropertyCmd = &cobra.Command{
...
+          },
+          "age": {
+              "type": "number"
+          }
+      }
+  }
+`
+
+func propertyDefinitionExample() string {
+	if !isYamlOutput() {
```
rocktavious, line 64: this var is populated by the flag `-o` and is only to be used when the end user does `-o yaml` on a `get` or `list` command.  It looks like you are using this function in the help text for a command.

This is conflating functionality and is not what `isYamlOutput` was intended for.  I'm not sure why we would show the `examplePropertyDefinitionInputFileJSON` in help text but you should beable to just put both examples into the help text if there is a reason for the JSON version.

IMO i would much rather we only support the `examplePropertyDefinitionInputFileYAML` and we throw an error when someone does the multiline string JSON version.

src/cmd/property.go
```diff
@@ -154,16 +201,16 @@ var updatePropertyDefinitionCmd = &cobra.Command{
 	Use:     "property-definition",
 	Aliases: []string{"propertydefinition", "propdef", "pd"},
 	Short:   "Update a property-definition",
-	Long:    `Update a property-definition`,
+	Long:    "Update a property-definition",
 	Example: fmt.Sprintf(`
-cat << EOF | opslevel update property-definition propdef3 -f -
+cat << EOF | opslevel create property-definition -f -
```
rocktavious, line 206: looks like this was erroneously renamed from `update` to `create` - and i think this also means removeal of the argument is incorrect since the code below still uses `args[0]` and `cobra.ExactArgs(1)`

src/cmd/property.go
```diff
@@ -176,42 +223,33 @@ EOF`, getYaml[opslevel.PropertyDefinitionInput]()),
...
 	}
+	return ReadResourceHandleJSONFields[opslevel.PropertyInput](input)
+}
 
-	if description, ok := data["description"].(string); ok {
-		propDefInput.Description = opslevel.RefOf(description)
-	}
-	if propertyDisplayStatus, ok := data["propertyDisplayStatus"].(string); ok {
-		propDefInput.PropertyDisplayStatus = opslevel.RefOf(opslevel.PropertyDisplayStatusEnum(propertyDisplayStatus))
+func ReadPropertyDefinitionInput(input []byte) (*opslevel.PropertyDefinitionInput, error) {
```
rocktavious, line 237: once again the `input` argument is only for testing and so it should not be in this function because its conflating 2 usecases with 1 logical function.  The tests should just use  `ReadResourceHandleJSONFields` and the code in production doesn't have to have `(nil)` everywhere

src/cmd/property.go
```diff
@@ -240,9 +278,10 @@ var listPropertyDefinitionsCmd = &cobra.Command{
 		if isJsonOutput() {
 			common.JsonPrint(json.MarshalIndent(list, "", "    "))
 		} else {
-			w := common.NewTabWriter("ALIASES", "ID", "NAME", "SCHEMA")
-			for _, item := range list {
-				fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", item.Aliases, item.Id, item.Name, item.Schema.ToJSON())
+			w := common.NewTabWriter("ALIASES", "ID", "NAME", "SCHEMA", "DISPLAY_STATUS", "ALLOWED_IN_CONFIG_FILES")
```
rocktavious, line 281: IMO we shouldn't print out the schema in the CSV it will make the table unwieldy.  If you feel strongly that we should then it should be the last column.

rocktavious (changes requested):

src/cmd/property.go
```diff
@@ -70,13 +116,15 @@ var listPropertyCmd = &cobra.Command{
 		if isJsonOutput() {
 			common.JsonPrint(json.MarshalIndent(properties.Nodes, "", "    "))
 		} else {
-			w := common.NewTabWriter("DEFINITION_ID", "VALUE", "LEN_VALIDATION_ERRORS")
-			for _, prop := range properties.Nodes {
+			w := common.NewTabWriter("DEF_ID", "ALIASES", "VALUE", "VALIDATION_ERRS", "LOCKED")
```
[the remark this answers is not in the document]
taimoorgit: I'll re-order the tables in #258

What did you mean by this part?

> Lastly what is the purpose of showing VALIDATION_ERRORS on a get -o csv command?

VALIDATION_ERRORS is just a variable that is part of the payload for property assignments, I didn't see a reason to exclude it.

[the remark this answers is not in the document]
taimoorgit: Still, I am going to exclude it, since it would rarely be useful.

PR #259, opened 2024-03-06
fix bug in list services maturity cmd

davidbloss:
## Issues

[263](https://github.com/OpsLevel/team-platform/issues/263)

## Changelog

Minor bugfix. Depends on [upstream opslevel-go fixes](https://github.com/OpsLevel/opslevel-go/pull/381)

- [X] List your changes here
- [ ] Make a `changie` entry

## Tophatting

```bash
# In orange acct, where we have over 100 services
opslevel list services maturity -o csv > service_maturity.csv
```

### DIFF
--- a/src/cmd/maturity.go
+++ b/src/cmd/maturity.go
@@ -30,7 +30,8 @@ There are multiple output formats that are useful
 		categoriesConn, err := client.ListCategories(nil)
 		cobra.CheckErr(err)
 		categories := categoriesConn.Nodes
-		data, err := client.ListServicesMaturity()
+		response, err := client.ListServicesMaturity(nil)
+		data := response.Nodes
 		cobra.CheckErr(err)
 		headers := []string{"Name", "Overall"}
 		sort.Slice(categories, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


src/cmd/maturity.go
```diff
@@ -30,7 +30,8 @@ There are multiple output formats that are useful
 		categoriesConn, err := client.ListCategories(nil)
 		cobra.CheckErr(err)
 		categories := categoriesConn.Nodes
-		data, err := client.ListServicesMaturity()
+		response, err := client.ListServicesMaturity(nil)
```
taimoorgit, line 33: @davidbloss the nilaway PR I have is not ready yet, but this is an example of a situation where we're vulnerable to nil pointer exceptions.

The `checkErr` should be done before accessing `response.Nodes` since `response` could be nil. 

Approving PR

taimoorgit (approved):

src/cmd/maturity.go
davidbloss: Shuffled ✅

PR #270, opened 2024-04-03
Add missing integration commands

rocktavious:
## Issues

Looks like we missed adding this command way back when.

## Changelog

- [ ] List your changes here
- [ ] Make a `changie` entry

## Tophatting

Still needs to be tested

### DIFF
--- a/src/cmd/integration.go
+++ b/src/cmd/integration.go
@@ -3,11 +3,10 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/mitchellh/mapstructure"
-
-	"github.com/opslevel/opslevel-go/v2024"
-
 	"github.com/opslevel/cli/common"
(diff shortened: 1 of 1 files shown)


src/cmd/integration.go
```diff
@@ -3,13 +3,73 @@ package cmd
...
+	input, err := readResourceInput[IntegrationInputType]()
+	if err != nil {
+		return output, err
+	}
+	if input.Version != CheckConfigCurrentVersion {
+		return output, fmt.Errorf("supported config version is '%s' but found '%s'",
+			IntegrationConfigCurrentVersion, input.Version)
+	}
+	// TODO: need to use input.Kind and a switch statement - but currently we only support AWS
+	if err := mapstructure.Decode(input.Spec, &output); err != nil {
```
rocktavious, line 43: this is the only part i'm unsure of and needs to be tested

davidbloss (approved):

PR #272, opened 2024-04-24
Kr/fix command

rocktavious:
## Issues

This fixes the `graph` command because whatever error checking was changed here completely breaks the command.  Not sure this was ever tested.


davidbloss (approved):
LGTM! 🎉

PR #301, opened 2024-07-10
add command to get service maturity on one service

davidbloss:
## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

```bash
opslevel get service maturity opslevel_cli
Name              Overall           📈 Quality         📋 Ownership       🔍 Observability   🔐 Security        🛠 Misc            🟢 Reliability
Opslevel CLI      💎 Platinum        💎 Platinum                          💎 Platinum        💎 Platinum
```


src/cmd/maturity.go
```diff
@@ -84,3 +80,39 @@ func GetValues(s *opslevel.ServiceMaturity, fields ...string) []string {
 	}
 	return output
 }
+
+func writeOutput(client *opslevel.Client, data []opslevel.ServiceMaturity) {
```
rocktavious, line 84: This is 😍  i've been meaning to do something like this in a more generic way but this is good because it starts to form the pattern.

src/cmd/list.go
```diff
@@ -22,7 +22,7 @@ func init() {
 }
 
 func isJsonOutput() bool {
-	return listOutputType == "json"
+	return listOutputType == "json" || getOutputType == "json"
```
rocktavious, line 25: This might break other `get *` commands.  We might need to not use `isJsonOutput` and instead move this logic into `writeOutput`

rocktavious (approved):

Issue #302, opened 2024-07-11
Regression with `update service` command


rocktavious:
A customer is running into a bug with CLI version `2024.5.28` when they try to use the `opslevel update service` command it sets all service summary values to none (with the exception of the owner field). They have moved to an older version in the meantime and the commands properly updates.

PR #305, opened 2024-07-18
opslevel update service command correctly updates fields

davidbloss:
## Issues

[Regression with update service command](https://github.com/OpsLevel/cli/issues/302)

## Changelog

Stop updating omitted fields (nil in this case) and only set fields to `null` if string is an empty string or `null`

- [X] List your changes here
- [X] Make a `changie` entry

### DIFF
--- a/.changes/unreleased/Bugfix-20240718-095431.yaml
+++ b/.changes/unreleased/Bugfix-20240718-095431.yaml
@@ -1,0 +1,3 @@
+kind: Bugfix
+body: opslevel update service command correctly updates fields
+time: 2024-07-18T09:54:31.699848-05:00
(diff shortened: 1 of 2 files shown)


rocktavious (approved):

rocktavious:
I know i had qualms about the `null` string being a thing.  But if this truely fixes something then i'm good with it.  I would be cautious about merging this though as it might have intended side effects because in json terms `"null"` is not the same as `null` but you are making them equal here

davidbloss:
> I know i had qualms about the `null` string being a thing. But if this truely fixes something then i'm good with it. I would be cautious about merging this though as it might have intended side effects because in json terms `"null"` is not the same as `null` but you are making them equal here

I took the "null" part out as it wasn't needed and if customers set something to `null` in json/yaml it's read in as `nil` anyway

PR #307, opened 2024-07-23
opslevel get and list service commands prints properties data

davidbloss:
## Issues

[Return service's custom properties in the json output](https://github.com/OpsLevel/team-platform/issues/427)

## Changelog

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

Run `opslevel get service sample_mobile_app`
```json
{
  "description": "Our mobile app. Find us in any App Store!",
  "framework": "Flutter",
  "htmlUrl": "https://app.opslevel.com/services/sample_mobile_app",
  "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMzMA",
  "aliases": [
    "sample_mobile_app"
  ],
  "language": "Dart",
  "name": "(Sample) Mobile App",
  # ... excluded some data
  "Dependencies": {
    "Edges": [
      {
        "Id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZURlcGVuZGVuY3kvMzE4MjE0",
        "Locked": false,
        "Node": {
          "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMyOA",
          "aliases": [
            "Shopping Cart Service",
            "ShoppingCartService",
            "cart",
            "sample_shopping_cart_service"
          ]
        },
        "Notes": ""
      }
    ],
    "PageInfo": {
      "HasNextPage": false,
      "HasPreviousPage": false,
      "Start": "MQ",
      "End": "MQ"
    }
  },
  "Dependents": {
    "Edges": null,
    "PageInfo": {
      "HasNextPage": false,
      "HasPreviousPage": false,
      "Start": "",
      "End": ""
    }
  },
  "Properties": {
    "Nodes": [
      {
        "Definition": {
          "id": "Z2lkOi8vb3BzbGV2ZWwvUHJvcGVydGllczo6RGVmaW5pdGlvbi8zOTY",
          "aliases": [
            "test_propdef"
          ]
        },
        "Locked": false,
        "Owner": {
          "OnService": {
            "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMzMA",
            "aliases": [
              "sample_mobile_app"
            ]
          }
        },
        "ValidationErrors": [],
        "Value": null
      }
    ],
    "PageInfo": {
      "HasNextPage": false,
      "HasPreviousPage": false,
      "Start": "MQ",
      "End": "MQ"
    },
    "TotalCount": 0
  }
}
```

Run `opslevel list service --properties --dependents --dependencies` (`-o json` omitted so we don't call for these extras)
```bash
NAME                            ID                                       ALIASES
(Sample) Inventory Tracker      Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMyOQ   sample_inventory_tracker
(Sample) Mobile App             Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMzMA   sample_mobile_app
(Sample) Shopping Cart Service  Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMyOA   Shopping Cart Service,ShoppingCartService,cart,sample_shopping_cart_service
example_name                    Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8xMjc2OTI  example_name
Temp GitLab CI Testing Service  Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8xMjM0NDM  temp_gitlab_ci_testing_service,tmp-gitlab-ci-testing-svc
Test Service One                Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8xMjU4ODA  test_service_one
```

### List services with `dependents` included
Run `opslevel list service -o json --dependencies | jq '.[] | select(.name == "(Sample) Mobile App") | .Dependencies'`
```json
{
  "Edges": [
    {
      "Id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZURlcGVuZGVuY3kvMzE4MjE0",
      "Locked": false,
      "Node": {
        "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMyOA",
        "aliases": [
          "Shopping Cart Service",
          "ShoppingCartService",
          "cart",
          "sample_shopping_cart_service"
        ]
      },
      "Notes": ""
    }
  ],
  "PageInfo": {
    "HasNextPage": false,
    "HasPreviousPage": false,
    "Start": "MQ",
    "End": "MQ"
  }
}```

### List services with `dependents` included
Run `opslevel list service -o json --dependents | jq '.[] | select(.name == "(Sample) Inventory Tracker") | .Dependents'`
```json
{
  "Edges": [
    {
      "Id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZURlcGVuZGVuY3kvMzE4MjE1",
      "Locked": false,
      "Node": {
        "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS85MzMyOA",
        "aliases": [
          "Shopping Cart Service",
          "ShoppingCartService",
          "cart",
          "sample_shopping_cart_service"
        ]
      },
      "Notes": ""
    }
  ],
  "PageInfo": {
    "HasNextPage": false,
    "HasPreviousPage": false,
    "Start": "MQ",
    "End": "MQ"
  }
}
```

### List services with `properties` included
Run `opslevel list service -o json --properties | jq '.[] | select(.name == "Test Service One") | .Properties'`
```json
{
  "Nodes": [
    {
      "Definition": {
        "id": "Z2lkOi8vb3BzbGV2ZWwvUHJvcGVydGllczo6RGVmaW5pdGlvbi8zOTY",
        "aliases": [
          "test_propdef"
        ]
      },
      "Locked": false,
      "Owner": {
        "OnService": {
          "id": "Z2lkOi8vb3BzbGV2ZWwvU2VydmljZS8xMjU4ODA",
          "aliases": [
            "test_service_one"
          ]
        }
      },
      "ValidationErrors": [],
      "Value": null
    }
  ],
  "PageInfo": {
    "HasNextPage": false,
    "HasPreviousPage": false,
    "Start": "MQ",
    "End": "MQ"
  },
  "TotalCount": 0
}
```

### DIFF
--- a/src/cmd/service.go
+++ b/src/cmd/service.go
@@ -217,9 +217,7 @@ func init() {
 }
 
 func hydrateService(service *opslevel.Service, client *opslevel.Client) {
-	_, err := service.GetDependencies(client, nil)
-	cobra.CheckErr(err)
-	_, err = service.GetDependents(client, nil)
+	_, err := service.GetDependents(client, nil)
 	cobra.CheckErr(err)
 	_, err = service.GetDependencies(client, nil)
 	cobra.CheckErr(err)
--- a/src/cmd/service.go
+++ b/src/cmd/service.go
@@ -69,12 +69,7 @@ var getServiceCmd = &cobra.Command{
 			service, err = getClientGQL().GetServiceWithAlias(key)
 			cobra.CheckErr(err)
 		}
-		_, err = service.GetDependents(client, nil)
-		cobra.CheckErr(err)
-		_, err = service.GetDependencies(client, nil)
-		cobra.CheckErr(err)
-		_, err = service.GetProperties(client, nil)
-		cobra.CheckErr(err)
+		hydrateService(service, client)
 		common.WasFound(service.Id == "", key)
 		common.PrettyPrint(service)
 	},
@@ -86,9 +81,14 @@ var listServiceCmd = &cobra.Command{
 	Short:   "Lists services",
 	Long:    `Lists services`,
 	Run: func(cmd *cobra.Command, args []string) {
-		resp, err := getClientGQL().ListServices(nil)
-		list := resp.Nodes
+		var list []opslevel.Service
+		client := getClientGQL()
+		resp, err := client.ListServices(nil)
 		cobra.CheckErr(err)
+		for _, service := range resp.Nodes {
+			hydrateService(&service, client)
+			list = append(list, service)
+		}
 		if isJsonOutput() {
 			common.JsonPrint(json.MarshalIndent(list, "", "    "))
 		} else if isCsvOutput() {
@@ -214,6 +214,17 @@ func init() {
 	deleteCmd.AddCommand(deleteServiceCmd)
 
 	importCmd.AddCommand(importServicesCmd)
+}
+
+func hydrateService(service *opslevel.Service, client *opslevel.Client) {
+	_, err := service.GetDependencies(client, nil)
+	cobra.CheckErr(err)
+	_, err = service.GetDependents(client, nil)
+	cobra.CheckErr(err)
+	_, err = service.GetDependencies(client, nil)
+	cobra.CheckErr(err)
+	_, err = service.GetProperties(client, nil)
+	cobra.CheckErr(err)
 }
 
 func convertServiceUpdateInput(input opslevel.ServiceUpdateInput) opslevel.ServiceUpdateInputV2 {
--- a/src/cmd/service.go
+++ b/src/cmd/service.go
@@ -73,6 +73,8 @@ var getServiceCmd = &cobra.Command{
 		cobra.CheckErr(err)
 		_, err = service.GetDependencies(client, nil)
 		cobra.CheckErr(err)
+		_, err = service.GetProperties(client, nil)
+		cobra.CheckErr(err)
 		common.WasFound(service.Id == "", key)
 		common.PrettyPrint(service)
 	},
--- a/src/cmd/service.go
+++ b/src/cmd/service.go
@@ -69,7 +69,16 @@ var getServiceCmd = &cobra.Command{
 			service, err = getClientGQL().GetServiceWithAlias(key)
 			cobra.CheckErr(err)
 		}
-		hydrateService(service, client)
+		// Extra fields only displayed in JSON format
+		if isJsonOutput() {
+			_, err = service.GetDependents(client, nil)
+			cobra.CheckErr(err)
+			_, err = service.GetDependencies(client, nil)
+			cobra.CheckErr(err)
+			_, err = service.GetProperties(client, nil)
+			cobra.CheckErr(err)
+		}
+
 		common.WasFound(service.Id == "", key)
 		common.PrettyPrint(service)
 	},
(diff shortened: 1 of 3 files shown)


taimoorgit (changes requested):
Looking at the tophatting I noticed that the field names are all capitalized like the struct names in Go. 

Issue 1 - We should update https://github.com/OpsLevel/opslevel-go/blob/main/service.go#L42 and also the connection type to have a tag like `json:"properties"`. This should be an easy fix

Issue 2 - Another problem is that we are returning page info about the properties despite having a `hydrateService` function. I think it was a mistake to have the Go struct for `Service` to leave property assignments as connection types. They should be arrays so that users do not suffer from this UX issue.

Issue 2 is something that we should solve because without it our tools offering for property assignments is broken and IMO the pagination info being there would be infuriating for a customer who is trying to be productive with the CLI when it comes to managing property assignments.

We should fix this and we (probably) have to introduce a breaking change to do so. We can discuss this in a standup/pairing session

image

Also, can you share what the output of a service with dependents and dependencies looks like?

src/cmd/service.go
```diff
@@ -214,6 +216,17 @@ func init() {
 	importCmd.AddCommand(importServicesCmd)
 }
 
+func hydrateService(service *opslevel.Service, client *opslevel.Client) {
+	_, err := service.GetDependencies(client, nil)
+	cobra.CheckErr(err)
+	_, err = service.GetDependents(client, nil)
+	cobra.CheckErr(err)
+	_, err = service.GetDependencies(client, nil)
```
taimoorgit, line 224: Function has been called twice

davidbloss: 🤦‍♂️ we just realllly need those dependencies 😬

rocktavious:
@taimoorgit while i hear your concerns - this is not our main objective right now and its my understanding this sweeping change to connection types would be a VERY large change that would balloon the scope significantly.  My understanding is this change came out of a request by a customer to just return the correct information.  I don't want to invest much time into this right now we have a HUGE backlog of things to be working on otherwise and this is one of the least important things.

If you both feel strongly that we shouldn't put this out the way it is and would rather do the bigger change then we need to tell the customer we cannot fix this right now and we'll schedule the bigger thing later in Q4 / Q1 next year.

taimoorgit:
@rocktavious @davidbloss So that we can ship the most minimal change to help the customer with their CLI issue,

When it comes to pagination, can we have the `hydrateService` function defined in this PR on the CLI end to automatically paginate all the property assignment `Nodes` so that the customer at least can get all property definitions in the response? Otherwise I think we are shipping an  unacceptable fix. 

And can we also do that for dependents/dependencies in that function? 

This way no opslevel-go changes are necessary and we don't have to worry about the overarching architecture of whether `Connnections` should live in resource `structs` which is a long rabbit hole.

rocktavious:
@taimoorgit what was originally how the "hydrate" style functions associated with service objects where suppose to work.  The reason we backed of from them was it hammered our API when you did `opslevel list services` because of the massive fan-out of requests.

The real solution is probably to add a wholly new command to get "all properties of 1 service"

taimoorgit (approved):
The `.Properties.TotalCount` shown in the output is very confusing since it is not accurate - customers could be mislead by assuming that a service has no properties because they checked the `.Properties.TotalCount` field and it was `0` despite the actual length of `.Properties.Nodes` being 2. 

I don't think that we are telegraphing this nearly enough, could we do something like print out a warning in coloured/bolded text to `/dev/stderr`?

The pagination data should also be removed in the future since it's not useful for CLI users.

For now this lets customers list relationships on a service whereas they could not before so it will unblock the customer that asked for this to be supported.

rocktavious:
If `TotalCount` is broken then thats on us.  we shouldn't be using total count from the API its not stable but we can calculate that ourselves within opslevel-go

rocktavious (approved):
Very clean - i like it!!!

PR #311, opened 2024-08-08
Db/integration tests

davidbloss:
## Issues

[CLI integration tests](https://github.com/OpsLevel/team-platform/issues/432)

## Changelog

Add integration tests on CRUD operations in cli. Creates, Updates, Deletes actual test resources

- [X] List your changes here
- [ ] Make a `changie` entry, N/A testing only

## Tophatting

`task test-integration`

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -50,7 +50,13 @@ tasks:
     desc: Run tests
     dir: "{{.SRC_DIR}}"
     cmds:
-      - go test -race -coverprofile=coverage.txt -covermode=atomic -v ./... {{ .CLI_ARGS }}
+      - go test -race -coverprofile=coverage.txt -covermode=atomic -v ./common/... {{ .CLI_ARGS }}
+
+  test-integration:
+    desc: Run Integration tests
+    dir: "{{.SRC_DIR}}"
+    cmds:
+      - go test -race -coverprofile=coverage.txt -covermode=atomic -v ./cmd/... {{ .CLI_ARGS }}
 
   update-opslevel-go:
     desc: Update opslevel-go version to latest release
--- a/src/cmd/team_test.go
+++ b/src/cmd/team_test.go
@@ -21,15 +21,13 @@ func Test_TeamCRUD(t *testing.T) {
 	}
 	teamId, err := createTeam(teamToCreate)
 	if err != nil {
-		t.Error(err)
-		t.FailNow()
+		t.Fatal(err)
 	}
 
 	// Get Team
 	createdTeam, err := getTeam(teamId)
 	if err != nil {
-		t.Error(err)
-		t.FailNow()
+		t.Fatal(err)
 	}
 	if createdTeam.Name != teamToCreate.Name ||
 		createdTeam.Responsibilities != *teamToCreate.Responsibilities {
@@ -44,8 +42,7 @@ func Test_TeamCRUD(t *testing.T) {
 	updatedTeamId, err := updateTeam(teamId, teamToUpdate)
 	if err != nil {
 		_ = deleteTeam(string(createdTeam.Id))
-		t.Error(err)
-		t.FailNow()
+		t.Fatal(err)
 	}
 	if string(createdTeam.Id) != updatedTeamId {
 		t.Errorf("Update 'team' failed, expected returned ID '%s' but got '%s'", string(createdTeam.Id), updatedTeamId)
--- a/src/cmd/team_test.go
+++ b/src/cmd/team_test.go
@@ -37,7 +37,10 @@ func Test_TeamCRUD(t *testing.T) {
 		t.Errorf("Get 'team' failed, got error: %v", err)
 	}
 
-	createdTeam := jsonToResource[opslevel.Team](getOutput)
+	createdTeam, err := jsonToResource[opslevel.Team](getOutput)
+	if err != nil {
+		t.Error("Failed to convert JSON from API to 'opslevel.Team' struct")
+	}
 	if createdTeam.Name != teamToCreate.Name ||
 		createdTeam.Responsibilities != *teamToCreate.Responsibilities {
 		t.Errorf("Create 'team' failed, expected team '%+v' but got '%+v'", teamToCreate, createdTeam)
--- a/src/cmd/team_test.go
+++ b/src/cmd/team_test.go
@@ -1,6 +1,7 @@
 package cmd_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/opslevel/cli/cmd"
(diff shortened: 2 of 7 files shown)


Taskfile.yml
```diff
@@ -50,7 +50,13 @@ tasks:
     desc: Run tests
     dir: "{{.SRC_DIR}}"
     cmds:
-      - go test -race -coverprofile=coverage.txt -covermode=atomic -v ./... {{ .CLI_ARGS }}
+      - go test -race -coverprofile=coverage.txt -covermode=atomic -v ./common/... {{ .CLI_ARGS }}
+
+  test-integration:
```
taimoorgit, line 55: I noticed that these 2 are the same, could add `test` as an alias for this task. https://taskfile.dev/usage/#task-aliases

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+// redirectStdout redirects os.Stdout to a pipe and returns the read and write ends of the pipe.
+func redirectStdout() (*os.File, *os.File) {
+	r, w, _ := os.Pipe()
+	oldStdout := os.Stdout
+	os.Stdout = w
+	return r, oldStdout
+}
+
+// captureOutput reads from r until EOF and returns the result as a string.
+func captureOutput(r *os.File, oldStdout *os.File) []byte {
```
taimoorgit, line 49: Function can be changed to return `([]byte, error)` so that we can handle the error on line 53 and also on line 36 since `execCmd` calls this function.

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+	w := os.Stdout
+	os.Stdout = oldStdout
+	w.Close()
+	gotOutput, _ := io.ReadAll(r)
+	return gotOutput
+	// return strings.TrimSpace(string(gotOutput))
```
taimoorgit, line 55: Remove this commented out line, this call is done on line 60

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// convert a simple API response to a string
+func asString(data []byte) string {
+	return strings.TrimSpace(string(data))
+}
+
+// convert JSON response from API to OpsLevel resource
+func jsonToResource[T any](jsonData []byte) *T {
```
taimoorgit, line 64: Also recommend that this function returns `(*T, error)` so that it can be handled by the calling test for example user_test.go line 37.

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+	var resource T
+	if err := json.Unmarshal(jsonData, &resource); err != nil {
+		return nil
+	}
+	return &resource
+}
+
+// write OpsLevel resource to YAML file for commands that read in a file
+func writeToYaml(userFileName string, opslevelResource any) error {
```
taimoorgit, line 73: Var should not be named `userFileName`, conflicts with the name of another package scoped variable.

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	rootCmd.SetArgs(cliArgs)
+	err := rootCmd.Execute()
+
+	output := captureOutput(r, oldStdout)
+	return output, err
+}
+
+// redirectStdout redirects os.Stdout to a pipe and returns the read and write ends of the pipe.
+func redirectStdout() (*os.File, *os.File) {
```
taimoorgit, line 41: Recommend that this change to returning `(*os.File, *os.File, error)` so that the error from `os.Pipe()` is handled.

src/cmd/team_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/opslevel/opslevel-go/v2024"
+)
+
+const (
+	teamFileName = "test_team.yaml"
+	teamName     = "CLI Test Team"
+)
+
+func Test_TeamCRUD(t *testing.T) {
```
taimoorgit, line 14: This would be much more readable if divided into functions like `Test_TeamCreate`, `Test_TeamUpdate` and so on, I would be able to quickly see what team related functions have tests by glancing at the functions defined in this file.

src/cmd/user_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/opslevel/opslevel-go/v2024"
+)
+
+const (
+	defaultUserRole = opslevel.UserRoleUser
+	userFileName    = "test_user.yaml"
+	userName        = "CLI Test User"
+)
+
+func Test_UserCRUD(t *testing.T) {
```
taimoorgit, line 16: Same here (separate test functions)

src/cmd/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+	Delete   Operation = "delete"
+	Example  Operation = "example"
+	Get      Operation = "get"
+	List     Operation = "list"
+	Update   Operation = "update"
+	Unassign Operation = "unassign"
+)
+
+// execute any OpsLevel CLI command
+func execCmd(cmd Operation, resource string, inputs ...string) ([]byte, error) {
```
taimoorgit, line 26: Could we do this? 

Rename `execCmd` to `ExecCmd` and keep it in the `cmd` package (since it needs access to `rootCmd`), and then remove everything else added in this PR a new `cmd_test` package?

This would give us better separation between CLI code and tests. However the CLI is not really a "library" we are providing so I don't hold this opinion strongly.

taimoorgit (changes requested):
Had some feedback,

1. We should handle all errors instead of ignoring them with `_`. If the error is really unusual for a test and it is tedious to have the calling function handle them, can we just `panic()` if `err != nil`?

2. Tests would be much more easy to read if each object had multiple `Test_` functions rather than only `Test_...CRUD` which does everything. 

3. We could add a `cmd_test` package (I don't hold this opinion strongly)

Taskfile.yml
davidbloss: The difference is `task test` runs tests in `./src/common/...` and `task test-integration` runs tests in `./src/cmd/...`

These should be separated as `task test` locally tests code and is part of `task ci`, whereas `task test-integration` is intended for pre-deploy integration testing and is intentionally kept out of regular testing

src/cmd/team_test.go
davidbloss: The code could be divided up into smaller functions but we can't run something like `Test_TeamUpdate` in isolation. We first need to create some test team to update

src/cmd/team_test.go
```diff
@@ -0,0 +1,106 @@
...
+	// Create Team
+	teamToCreate := opslevel.TeamCreateInput{
+		Name:             teamName,
+		Responsibilities: opslevel.RefOf("all the things"),
+	}
+	teamId, err := createTeam(teamToCreate)
+	if err != nil {
+		t.Error(err)
```
taimoorgit, line 24: Request that we replace these `t.Error(err)` and `t.FailNow()` calls with `t.Fatal(err)`

https://pkg.go.dev/testing#T.FailNow

>Error is equivalent to Log followed by Fail.

>Fatal is equivalent to Log followed by FailNow.

Therefore `t.Fatal(err)` is the better call to make.

src/cmd/team_test.go
```diff
@@ -0,0 +1,106 @@
...
+		t.FailNow()
+	}
+
+	// Get Team
+	createdTeam, err := getTeam(teamId)
```
taimoorgit, line 29: When it comes to all of these CRUD functions that the code has been divided into, we could actually pass in `t *testing.T` into them, and have those functions call `t.Fatal(err)` so that we don't have to examine return values and this function could be much shorter.

taimoorgit (changes requested):
Great revision, I have another recommendation to consider before merging

src/cmd/team_test.go
davidbloss: There are problems with automatically failing in each function.

If `updateTeam(...)` fails to update the created test team, we need to the flexibility to delete the team outside of that function. Other tests down the road might make user of `updateTeam()` or even `getTeam()` and those tests would need to determine how to handle those situations. We might want to run `getTeam()` to prove a team has been deleted for instance.

taimoorgit (approved):

Issue #304, opened 2024-07-15
Unable to update / create tool links on a service via the CLI


rocktavious:
https://jk-labs.slack.com/archives/C0296S8KJJ3/p1720717440253859

It currently appears impossible to create or update tool links on a service in the CLI.  We should add this functionality.

PR #313, opened 2024-08-12
add new command to create, update, and delete a Service Tool

davidbloss:
## Issues

[Create / Update tool links on a service](https://github.com/OpsLevel/cli/issues/304)

## Changelog

Add new commands to create, update, and delete a Service Tool

- [X] List your changes here
- [X] Make a `changie` entry

## Tophatting

Given this `new_tool.yaml` input file:
```yaml
displayName: "test tool"
category: deployment
environment: "dev"
url: "https://example-test.com"
```

### Create service tool, `opslevel create service tool service-alias -f new_tool.yaml`
```bash
# Tool ID is printed
"Z2lkOi8vb3BzbGV2ZWwvVG9vbC83MjkyMjM"
```

Update this `new_tool.yaml` input file:
```yaml
displayName: "Spiffy test tool"
category: production
environment: "prod"
url: "https://example-prod.com"
```

### Update service tool, `opslevel update service tool Z2lkOi8vb3BzbGV2ZWwvVG9vbC83MjkyMjM -f new_tool.yaml`
```bash
"Z2lkOi8vb3BzbGV2ZWwvVG9vbC83MjkyMjM"
```

### Delete service tool, `opslevel delete service tool "Z2lkOi8vb3BzbGV2ZWwvVG9vbC83MjkyMjM"`
```bash
deleted 'Z2lkOi8vb3BzbGV2ZWwvVG9vbC83MjkyMjM' service tool
```


src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+service: my-service-alias
+category: deployment
+displayName: "fancy tool"
+environment: "dev"
+url: "https://example.com"
+EOF
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getClientGQL()
+		serviceIdentifier := readServiceFieldFromYaml()
```
rocktavious, line 27: what about putting it as an `arg` to the command.  I think we do this in other places too - so it would be like this
```
cat << EOF | opslevel create service tool my-service-alias -f -
category: deployment
displayName: "fancy tool"
environment: "dev"
url: "https://example.com"
EOF
```

This would eliminate all that extra reading from yaml stuff

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+		cobra.CheckErr(err)
+
+		err = verifyToolCategory(toolFromInput.Category)
+		cobra.CheckErr(err)
+
+		client := getClientGQL()
+		service, err := getService(serviceIdentifier)
+		cobra.CheckErr(err)
+
+		if !isToolIdInServiceTools(toolId, service.Tools) {
```
rocktavious, line 76: This feels like unnecessary validation - the API will do this for us AND this incurs extra API calls - service get which we just found out is heavy

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	for _, serviceTool := range serviceTools.Nodes {
+		if toolId == serviceTool.Id {
+			return true
+		}
+	}
+	return false
+}
+
+func verifyToolCategory(toolCategory *opslevel.ToolCategory) error {
```
rocktavious, line 98: is the API error for this one bad?  if its not then we should remove this.  I really want to rely upon API as much as possible because validation code like this can easily go stale.  Say for example a new tool category is added now our CLI rejects it where as before it would just work.  Better forward compatibility of older CLI's for the future.

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+		"'category:' field has value '%s' but must be one of '%v'",
+		string(*toolCategory),
+		strings.Join(opslevel.AllToolCategory, ", "),
+	)
+}
+
+var deleteServiceToolCmd = &cobra.Command{
+	Use:        "tool ID",
+	Short:      "Delete a service tool",
+	Example:    `opslevel delete service tool <tool-ID> `,
```
rocktavious, line 112:
```suggestion
	Example:    `opslevel delete service tool <tool-ID>`,
```

src/cmd/service.go
```diff
@@ -284,3 +286,28 @@ func NullableString(value *string) *opslevel.Nullable[string] {
 	}
 	return opslevel.NewNullableFrom(*value)
 }
+
+func getService(identifier string) (*opslevel.Service, error) {
```
rocktavious, line 290: I swear we had something like this in opslevel-go.  If not we should add this to opslevel-go and/or modify `GetService` to work like this.

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+var updateServiceToolCmd = &cobra.Command{
+	Use:     "tool",
+	Aliases: []string{"tool"},
+	Short:   "Update service tool",
+	Args:    cobra.ExactArgs(1),
+	Example: `
+cat << EOF | opslevel update service tool tool-ID -f -
+service: my-service-alias
```
rocktavious, line 51: this field shouldn't be needed since you have the "ID" of the tool to update right?

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return fmt.Errorf(
+		"'category:' field has value '%s' but must be one of '%v'",
+		string(*toolCategory),
+		strings.Join(opslevel.AllToolCategory, ", "),
+	)
+}
+
+var deleteServiceToolCmd = &cobra.Command{
+	Use:        "tool ID",
```
rocktavious, line 110:
```suggestion
	Use:        "tool",
```

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+
+		toolFromInput.ServiceId = &service.Id
+		tool, err := client.CreateTool(*toolFromInput)
+		cobra.CheckErr(err)
+		common.PrettyPrint(string(tool.Id))
+	},
+}
+
+var updateServiceToolCmd = &cobra.Command{
+	Use:     "tool",
```
rocktavious, line 45:
```suggestion
	Use:     "tool TOOL-ID",
	ArgAliases: []string{"TOOL-ID"},
```

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+	Aliases: []string{"tool"},
```
rocktavious, line 46:
```suggestion
```

src/cmd/tools.go
```diff
@@ -0,0 +1,131 @@
...
+		strings.Join(opslevel.AllToolCategory, ", "),
+	)
+}
+
+var deleteServiceToolCmd = &cobra.Command{
+	Use:        "tool ID",
+	Short:      "Delete a service tool",
+	Example:    `opslevel delete service tool <tool-ID> `,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
```
rocktavious, line 114:
```suggestion
```

rocktavious (changes requested):
Overall gets the job done - 1 major question about the customer interface for targeting the service and then a bunch of little fixes.

src/cmd/service.go
davidbloss: There is no catch all like this in opslevel-go. Something like `GetServiceWithIdentifier(identifier string)` then mark the others as deprecated?

rocktavious: I think we can modify `GetService` to accept a string - ID is a type alias for string IIRC - and then have it work if the string passed in is an ID or Alias.  Would need to test a before and after but i think thats a safe non-breaking interface change to the `GetService` method.

src/cmd/tools.go
davidbloss: 😬 yes this needs to go. I think this was a product of my fevered thinking, literally

src/cmd/tools.go
davidbloss: Good call out. Adding this in the next commit

src/cmd/tools.go
davidbloss: Correct, `service:` removed in the next commit

src/cmd/service.go
davidbloss: I tried a few things and could not get this to work

src/cmd/tools.go
```diff
@@ -0,0 +1,91 @@
...
+url: "https://example.com"
+EOF
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		serviceAlias := args[0]
+		toolFromInput, err := readResourceInput[opslevel.ToolCreateInput]()
+		cobra.CheckErr(err)
+
+		client := getClientGQL()
+		serviceId, err := client.GetServiceIdWithAlias(serviceAlias)
```
rocktavious, line 29: shouldn't this use `getService`?

src/cmd/tools.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+var updateServiceToolCmd = &cobra.Command{
+	Use:        "tool TOOL-ID",
+	ArgAliases: []string{"TOOL-ID"},
+	Short:      "Update service tool",
+	Args:       cobra.ExactArgs(1),
+	Example: `
+cat << EOF | opslevel update service tool tool-ID -f -
+service: my-service-alias
```
rocktavious, line 46: I think we were suppose to remove this field from the example.

src/cmd/tools.go
```diff
@@ -0,0 +1,91 @@
...
+environment: "dev"
+url: "https://example.com"
+EOF
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		id := args[0]
+		if !opslevel.IsID(id) {
+			cobra.CheckErr(fmt.Errorf("invalid ID: '%s'", id))
+		}
+		toolId := opslevel.ID(id)
```
rocktavious, line 58:
```suggestion
``

File: src/cmd/tools.go [hunk 14]
```diff
@@ -0,0 +1,91 @@
...
+		toolFromInput, err := readResourceInput[opslevel.ToolUpdateInput]()
+		cobra.CheckErr(err)
+
+		toolFromInput.Id = toolId
```
rocktavious, line 62:
```suggestion
		toolFromInput.Id = opslevel.ID(id)

```

rocktavious (changes requested):
Still a few more issues

rocktavious (approved):
LGTM 🚀

PR #336, opened 2024-10-08
added new reactivate integration command

davidbloss:
Resolves [#506](https://github.com/OpsLevel/team-platform/issues/506)

### Checklist

### DIFF
--- a/src/cmd/integration.go
+++ b/src/cmd/integration.go
@@ -240,7 +240,7 @@ var reactivateIntegrationCmd = &cobra.Command{
 		integration, err := getClientGQL().IntegrationReactivate(args[0])
 		cobra.CheckErr(err)
 
-		fmt.Printf("reactivated integration with id '%s'\n", integration.Id)
+		fmt.Printf("reactivated integration '%s' with id '%s'", integration.Name, integration.Id)
 	},
 }
 
--- a/src/cmd/integration.go
+++ b/src/cmd/integration.go
@@ -229,9 +229,26 @@ EOF
 	},
 }
 
+var reactivateIntegrationCmd = &cobra.Command{
+	Use:        "reactivate ID",
+	Short:      "Reactivate an integration",
+	Long:       `Reactivate an integration that was invalidated or deactivated`,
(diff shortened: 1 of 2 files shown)


src/cmd/integration.go
```diff
@@ -229,9 +229,26 @@ EOF
 	},
 }
 
+var reactivateIntegrationCmd = &cobra.Command{
+	Use:        "reactivate ID",
```
rocktavious, line 233:
```suggestion
	Use:        "reactivate ID|ALIAS",
```

src/cmd/integration.go
```diff
@@ -229,9 +229,26 @@ EOF
...
+	Short:      "Reactivate an integration",
+	Long:       `Reactivate an integration that was invalidated or deactivated`,
+	Example:    `opslevel update integration reactivate Z2lkOi8vb123456789`,
+	Args:       cobra.ExactArgs(1),
+	ArgAliases: []string{"ID"},
+	Run: func(cmd *cobra.Command, args []string) {
+		integration, err := getClientGQL().IntegrationReactivate(args[0])
+		cobra.CheckErr(err)
+
+		fmt.Printf("reactivated integration with id '%s'\n", integration.Id)
```
rocktavious, line 243: Do we use this in other places or should we use the `log.Info...` way of outputting to the end user?

src/cmd/integration.go
davidbloss: I actually started with this then found out that integrations don't have aliases

src/cmd/integration.go
davidbloss: 🤔 I see a lot more instances of `fmt.Printf` than `log.Info`.

Totally, I'll add the integration name 🚀

src/cmd/integration.go
rocktavious: Ahh 😢 

But they still support `IdentifierInput` on the way in?

src/cmd/integration.go
rocktavious: 👍

rocktavious (approved):

PR #352, opened 2024-11-12
Import a csv file to delete users + count flag for services

tozach:
Resolves #

### Problem

1. The CLI doesnt have a `bulk delete user`s feature. Additional scripting is required to delete multiple users
2. Need a simple solution to get the count of services than list the services and then count it

### Solution

Added `import user delete` in the CLI to support deletion of multiple users by email
Added `count` flag to get a count of the services

### Checklist

- [ -] I have run this code, and it appears to resolve the stated issue.

### DIFF
--- a/src/cmd/user.go
+++ b/src/cmd/user.go
@@ -99,25 +99,19 @@ var listUserCmd = &cobra.Command{
 	Example: `
 opslevel list user
 opslevel list user --ignore-deactivated
-opslevel list user --deactivated
 opslevel list user -o json | jq 'map({"key": .Name, "value": .Role}) | from_entries'
 `,
 	Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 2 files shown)


src/cmd/user.go
```diff
@@ -215,6 +215,41 @@ EOF
...
 }
 
+var bulkDeleteUsersCmd = &cobra.Command{
+	Use:     "delete users",
+	Aliases: []string{"bulk delete users"},
+	Short:   "Deletes users from a CSV",
+	Long: `Deletes a list of users from a CSV file with the column headers:
+Email`,
+	Example: `
+cat << EOF | opslevel import delete users -f -
```
rocktavious, line 225: @tozach - so I think the where we want to put this is in the normal `delete users` command but add a `--bulk` or  `--csv`.  Thats a bit more of a surgery so do you mind if I checkout this branch and muck with the code to get that to work?  

I definitely don't want it under the `import` command

PR #364, opened 2024-12-13
Update install.sh

rocktavious:
Resolves #

### Problem

There exists an edgecase where `root is not in the sudoers file` .

### Solution

so this change tries to mitigate that by looping through the PATH and trying to find a writeable directory we can install to.

### Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -88,7 +88,7 @@ install_cli() {
         echo "Installing the OpsLevel CLI to '$TARGET_DIR' ..."
     fi
 
-    mv "$TMP_DIR/opslevel" /usr/local/bin/
+    mv "$TMP_DIR/opslevel" "$TARGET_DIR"
     if [ $? -ne 0 ]; then
         echo "Installation failed."
         exit 1
--- a/install.sh
+++ b/install.sh
@@ -46,6 +46,11 @@ get_version() {
     fi
 }
 
+# Function to check if a directory is writable
+is_writable() {
+    [ -d "$1" ] && [ -w "$1" ]
+}
+
 # Function to download and install the CLI tool
 install_cli() {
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -64,9 +69,26 @@ install_cli() {
...
+
+    # If no writable directory is found, exit
+    if [ -z "$TARGET_DIR" ]; then
+        echo "Installation failed.  User has no permissions to any directory on PATH"
+        exit 1
+    else
+        echo "Installing the OpsLevel CLI to '$TARGET_DIR' ..."
+    fi
+
+    mv "$TMP_DIR/opslevel" /usr/local/bin/
```
rocktavious, line 91:
```suggestion
    mv "$TMP_DIR/opslevel" "$TARGET_DIR"
```

install.sh
```diff
@@ -64,9 +69,26 @@ install_cli() {
...
+    for dir in $(echo "$PATH" | tr ':' '\n'); do
+        if is_writable "$dir"; then
+            TARGET_DIR="$dir"
+            break
+        fi
+    done
+
+    # If no writable directory is found, exit
+    if [ -z "$TARGET_DIR" ]; then
+        echo "Installation failed.  User has no permissions to any directory on PATH"
```
rocktavious, line 85:
```suggestion
    if [ -z "$TARGET_DIR" ]; then
        if [ -n "${GITHUB_PATH:-}" ] && [ -w "$GITHUB_PATH" ]; then
            TARGET_DIR="$HOME/.local/bin"
            mkdir -p "$TARGET_DIR"
            echo "$TARGET_DIR" >> "$GITHUB_PATH"
            echo "No writable directory in PATH. Using GITHUB_PATH to add $TARGET_DIR to PATH."
        else
            echo "Installation failed.  User has no permissions to any directory on PATH"
        fi
```

rocktavious: Going to actually move this into the action so it ensures a writable directory is on the path

davidbloss (approved):

PR #385, opened 2025-03-27
Update .goreleaser.yml

rocktavious:
Resolves #

### Problem

https://goreleaser.com/deprecations/#archivesformat_overridesformat

### Solution

Perform syntax upgrade

### Checklist

- [x] Make a [changie](https://github.com/OpsLevel/cli/blob/main/CONTRIBUTING.md#changie-change-log-generation) entry that explains the customer facing outcome of this change

### DIFF
--- a/src/.goreleaser.yml
+++ b/src/.goreleaser.yml
@@ -64,7 +64,7 @@ archives:
   - name_template: '{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}'
     format_overrides:
       - goos: windows
-        format: zip
+        format: [ 'zip' ]
 brews:
   - name: cli
     homepage: "https://www.opslevel.com/"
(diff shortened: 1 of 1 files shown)


src/.goreleaser.yml
```diff
@@ -64,7 +64,7 @@ archives:
   - name_template: '{{ .ProjectName }}-{{ .Os }}-{{ .Arch }}'
     format_overrides:
       - goos: windows
-        format: zip
+        format: [ 'zip' ]
```
derek-etherton-opslevel, line 67: The deprecation was [changing the key from format to formats](https://goreleaser.com/deprecations/#archivesformat). Apparently single string should still be fine?

> It will still accept a single string, e.g.: formats: zip. In most cases you can simply rename the property to formats.

```suggestion
        formats: zip
```

derek-etherton-opslevel (approved):

PR #389, opened 2025-04-15
Kr/mcp add server mode

rocktavious:
Resolves #

### Problem

MCP servers can run in 2 modes - stdio and http

### Solution

Expose commandline arguments to configure the mode and the port for `http` mode

### Checklist

- [ ] ~Make a [changie](https://github.com/OpsLevel/cli/blob/main/CONTRIBUTING.md#changie-change-log-generation) entry that explains the customer facing outcome of this change~


eapache-opslevel (approved):

eapache-opslevel:
Do we need to do any of the items in https://modelcontextprotocol.io/docs/concepts/transports#security-warning%3A-dns-rebinding-attacks to protect the security of this mode?

eapache-opslevel:
We are going to let this linger a bit. We don't understand the security implications fully, and all anybody needs right now is the STDIO version which has no security concerns.

rocktavious:
Additionally  `client := getClientGQL()` is reading the API token from env vars not from headers passed over the wire.  So we'll need to modify that aspect when running in HTTP mode.

rocktavious:
Won't Do.  We've decided to switch libraries so this will have to be slightly re-written.  going to close for now but keep the branch for when we want SSE with `mcp-go`

PR #390, opened 2025-04-15
Add Domains and Systems

rocktavious:
Resolves #

### Problem

The current MCP cannot find information about your systems and domains in opslevel

### Solution

Register tools for these objects

### Checklist


BigBlue79 (approved):
Validated with the MCP inspector locally using a prod API token 👍

PR #394, opened 2025-04-16
drop --remote flag from submodule setup, use --recursive Just In Case

wesleyjellis:
Resolves #

### Problem

`task setup` will checkout/update the submodule to the latest commit on the remote branch (I'm guessing main) instead of what's been "pinned" / set 

### Solution

Don't use `--remote` [docs](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---remote)

### Checklist

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -68,7 +68,7 @@ tasks:
     cmds:
       - cmd: echo "Setting up opslevel-go workspace..."
         silent: true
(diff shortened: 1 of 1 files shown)


rocktavious (approved):

eapache-opslevel:
🎉 I wondered why this had been happening

PR #400, opened 2025-04-17
Add 'mcp' to mcp graphql UA

eapache-opslevel:
This requires creating a separate graphql client, but we only do that if we are running the MCP server so the overhead is minimal.


rocktavious (approved):
🚀

PR #402, opened 2025-05-30
Kr/upgrade opslevel go 2025

rocktavious:
Resolves #

### Problem

Untill now the CLI has been on opslevel-go 2024 because of a huge nullability shift along with a number of struct interface changes due to code-gen and standardization.  This has prevented us from upgrading the CLI.

### Solution

Paydown this tech debt and make it so the CLI is on the latest opslevel-go release.

### Checklist

### DIFF
--- a/src/cmd/property.go
+++ b/src/cmd/property.go
@@ -61,11 +61,7 @@ var listPropertyCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		var service *opslevel.Service
 		var err error
-		if opslevel.IsID(args[0]) {
-			service, err = getClientGQL().GetService(*opslevel.NewID(args[0]))
-		} else {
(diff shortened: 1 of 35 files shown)


src/cmd/property.go
```diff
@@ -61,11 +61,7 @@ var listPropertyCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		var service *opslevel.Service
 		var err error
-		if opslevel.IsID(args[0]) {
-			service, err = getClientGQL().GetService(*opslevel.NewID(args[0]))
-		} else {
-			service, err = getClientGQL().GetServiceWithAlias(args[0])
-		}
+		service, err = getClientGQL().GetService(args[0])
```
derek-etherton-opslevel, line 64: seeing some nice cleanups from the upgrade 👍

PR #404, opened 2025-06-12
Setup e2e tests for some resources

rocktavious:
Resolves #

### Problem

There is no way to know if we are creating regressions with this 2025 upgrade without extensive manual testing back and forth between current and older versions of the CLI

### Solution

Add a nice pattern for creating end to end (e2e) tests for the CLI so this can be done easily

### Checklist

- [x] Make a [changie](https://github.com/OpsLevel/cli/blob/main/CONTRIBUTING.md#changie-change-log-generation) entry that explains the customer facing outcome of this change


jasonopslevel (approved):
Looks like a good start to e2e tests, ran locally and looks good!

